PR #5, opened 2019-11-29
Spherical to geodetic and inverse transformations

leouieda:
Add methods to Ellipsoid to transform coordinates from geodetic to
spherical and back again. Copied from Harmonica. Added as methods to
avoid passing "ellipsoid" all the time or having to use a global
ellipsoid.

**Reminders**

- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.

### DIFF
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -145,3 +145,105 @@ class Ellipsoid:
             self.geocentric_grav_const * (1 + self.emm * aux) / self.semimajor_axis ** 2
         )
         return result
+
+    def geodetic_to_spherical(self, longitude, latitude, height):
+        """
+        Convert from geodetic to geocentric spherical coordinates.
+
+        The geodetic datum is defined by this ellipsoid. The coordinates are
+        converted following [Vermeille2002]_.
+
+        Parameters
+        ----------
+        longitude : array
+            Longitude coordinates on geodetic coordinate system in degrees.
+        latitude : array
+            Latitude coordinates on geodetic coordinate system in degrees.
+        height : array
+            Ellipsoidal heights in meters.
+
+        Returns
+        -------
+        longitude : array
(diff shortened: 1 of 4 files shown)


leouieda:
@santisoler as I started thinking about this package, it feels like a fancy global ellipsoid mechanism is more work than necessary. In the end, the only things that it would be used for in Harmonica is the normal gravity. It probably just makes more sense to have these as methods of an ellipsoid. Then setting the ellipsoid is just instantiating the right class (WGS84, Mars, etc). 

That kind of makes sense for the coordinate conversions. And if we're doing that, it might make more sense to have the normal gravity here as well (as a method).

What  do you think?

santisoler:
I'm a little bit overwhelmed with all the new packages and the goals for them. I do like having separate libraries for different tasks, but I don't want to have a large number of packages that do very little stuff. It's very clear that Verde has Greens' functions interpolators and Harmonica has functions for processing and forward modeling gravity and magnetics. 

Verde is also handling coordinates, but Harmonica provides some coordinates conversions. Now Boule will handle ellipsoids, but also coordinates conversions and (maybe) normal gravity calculation.

I think we are scattering coordinates management among all packages and probably loosing (or mutating) the goals for our packages.

I like the fancy ellipsoid interface because it prevents the possible mistake of using a wrong ellipsoid, and also lowers the number of arguments needed to be passed to some functions. For example, for the coordinates conversions we only need to pass the coordinates. But if we instantiate an Ellipsoid class we would need to pass it to the conversion function.

I don't have a robust solution for these things, but I think we should start thinking about them before continue moving staff from one package to another. Am I right?

leouieda:
Summarizing our discussions on this:

1. It makes sense to keep Boule to foster collaboration with other projects
2. Integrating these things into the classes is simpler and easier to maintain than the hack of the global ellipsoid variable

PR #13, opened 2020-01-09
WIP Add an ellipsoid for the Moon

leouieda:
Use values from the compilation in [Wieczorek (2015)](https://doi.org/10.1016/B978-0-444-53802-4.00169-X).

**Reminders**

- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.


leouieda:
Tests fail because the equations fail (nan) when the flattening is 0. That leads to eccentricities being 0 and there are divisions of these values in the equations. The only sensible way forward I can see is to create a separate `Sphere` class that has simplified equations for a reference sphere. It would actually be a lot simpler since we'd be using spherical coordinates by default and so wouldn't need the coordinate conversions. The gravity equations is also a lot simpler.

leouieda:
Maybe a better way forward is to check if the flattening is close to zero and then use the sphere equation instead. That way we don’t need a new class and the associated maintenance.

santisoler:
I have updated this PR by defining the `MOON` ellipsoids as an instance of the new `Sphere` class introduced on #42.
I faced a bug on how the `Sphere` class computes the gravity on the equator, which should be fixed on #48.
After that, I think the Moon implementation should be ready, maybe we could add a gallery example for it.

leouieda:
Closing this since the approach wouldn't work. See #53 instead.

PR #35, opened 2020-07-08
Add method for computing prime vertical radius

santisoler:
Add private method to `boule.Ellipsoid` for computing the prime vertical radius.
The definition of the prime vertical radius can be seen on [Vermeille (2002)](https://doi.org/10.1007/s00190-002-0273-6)  and [Vajda (2004)](https://d1wqtxts1xzle7.cloudfront.net/45194692/On_evaluation_of_Newton_integrals_in_geo20160429-31556-16b23iv.pdf?1461926098=&response-content-disposition=inline%3B+filename%3DOn_evaluation_of_Newton_integrals_in_geo.pdf&Expires=1594246043&Signature=G0FRusq9DVc1eSllnrFGZUufrskJdDQgXpOn8pzITqSQVvnwflK2a9nSZqUbfqD2msqsDEVbUiaNeUKEeXYentp3GiMmEXE~P-Dt9KddO0-AcaXf-WjTUWeJb5wmnFKFAFANWhNS4x9IlVY2NdqPP2U-Ide6KxA7yNGN-JhQ1i-OvQEG8dSAgvYJTDKhdodlWcLmn1zYtlDGtde3IMEWT2JpB-JzoXRF8aJbEMTKXEGvnb4FMgGxWsfsCmFry67Zu3pAebeRrvGhY-JYuODa2bGqr8nu8CyO0KuOV5eoVFLI2zFZWIXwDH1njr~2kLhvJwBMm9q4r7joTRAPvnmcbw__&Key-Pair-Id=APKAJLOHF5GGSLRBV4ZA).
It's usually symbolized with the letter N and depends on the geometry of the ellipsoid and it must be evaluated on a certain latitude angle.
We were already computing it for applying the geodetic to geodetic to spherical coordinate conversion.
It is also needed for computing the Euclidean distance between two points given in geodetic coordinates (see Vajda, 2004), feature that will be added after solving https://github.com/fatiando/harmonica/issues/154 .
In order to prevent rewriting the equation, it's better to have a private function inside `boule.Ellipsoid`.
The new method takes the precomputed sine of the latitude angle to make it more efficient.

**Reminders**:

### DIFF
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -146,12 +146,11 @@ class Ellipsoid:
         )
         return result
 
-    def _prime_vertical_radius(self, latitude_radians):
-        """
+    def _prime_vertical_radius(self, sinlat):
+        r"""
         Calculate the prime vertical radius for a given latitude
 
         The prime vertical radius is compute following:
-
 
         .. math::
 
@@ -161,16 +160,16 @@ class Ellipsoid:
 
         Parameters
         ----------
-        latitude_radians : float or array-like
-            Latitude values given in radians.
+        sinlat : float or array-like
+            Sine of the latitude angle.
 
         Returns
         -------
         prime_vertical_radius : float or array-like
-            Prime vertical radius given in the same units as the semimajor axis.
+            Prime vertical radius given in the same units as the semimajor axis
         """
         return self.semimajor_axis / np.sqrt(
-            1 - self.first_eccentricity ** 2 * np.sin(latitude_radians) ** 2
+            1 - self.first_eccentricity ** 2 * sinlat ** 2
         )
 
     def geodetic_to_spherical(self, longitude, latitude, height):
@@ -203,13 +202,14 @@ class Ellipsoid:
 
         """
         latitude_rad = np.radians(latitude)
-        prime_vertical_radius = self._prime_vertical_radius(latitude_rad)
+        coslat, sinlat = np.cos(latitude_rad), np.sin(latitude_rad)
+        prime_vertical_radius = self._prime_vertical_radius(sinlat)
         # Instead of computing X and Y, we only compute the projection on the
         # XY plane: xy_projection = sqrt( X**2 + Y**2 )
-        xy_projection = (height + prime_vertical_radius) * np.cos(latitude_rad)
+        xy_projection = (height + prime_vertical_radius) * coslat
         z_cartesian = (
             height + (1 - self.first_eccentricity ** 2) * prime_vertical_radius
-        ) * np.sin(latitude_rad)
+        ) * sinlat
         radius = np.sqrt(xy_projection ** 2 + z_cartesian ** 2)
         spherical_latitude = np.degrees(np.arcsin(z_cartesian / radius))
         return longitude, spherical_latitude, radius
@@ -289,7 +289,7 @@ class Ellipsoid:
             The (geodetic) latitude where the normal gravity will be computed
             (in degrees).
         height : float or array
-            The ellipsoidal (geometric) height of computation point (in meters).
+            The ellipsoidal (geometric) height of computation point (in meters)
 
         Returns
         -------
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -146,6 +146,33 @@ class Ellipsoid:
         )
         return result
 
+    def _prime_vertical_radius(self, latitude_radians):
+        """
+        Calculate the prime vertical radius for a given latitude
+
+        The prime vertical radius is compute following:
+
+
+        .. math::
+
+            N(\phi) = \frac{a}{\sqrt{1 - e^2 \sin^2(\phi)}}
+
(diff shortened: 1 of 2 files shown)


santisoler:
@leouieda Please take a look at this when you have time.

I decided it's effectively better to have this method. Because the method it's private, I took the liberty of setting the sine of the latitude angle as argument, so it can be efficiently used for coordinates conversions (in Boule) and for calculating geodetic distances (in Harmonica).

boule/ellipsoid.py
```diff
@@ -146,6 +146,32 @@ def gravity_pole(self):
         )
         return result
 
+    def _prime_vertical_radius(self, sinlat):
+        r"""
+        Calculate the prime vertical radius for a given latitude
```
leouieda, line 151:
```suggestion
        Calculate the prime vertical radius for a given geodetic latitude
```

boule/ellipsoid.py
```diff
@@ -146,6 +146,32 @@ def gravity_pole(self):
...
+
+        The prime vertical radius is compute following:
```
leouieda, line 153:
```suggestion
        The prime vertical radius is defined as:
```

boule/ellipsoid.py
```diff
@@ -146,6 +146,32 @@ def gravity_pole(self):
...
+
+        .. math::
+
+            N(\phi) = \frac{a}{\sqrt{1 - e^2 \sin^2(\phi)}}
+
+        Where $a$ is the semimajor axis and $e$ the first eccentricity.
```
leouieda, line 159:
```suggestion
        Where :math:`a` is the semimajor axis and :math:`e` is the first eccentricity.
```

boule/ellipsoid.py
```diff
@@ -146,6 +146,32 @@ def gravity_pole(self):
...
+
```
leouieda, line 160:
```suggestion

        This function receives the sine of the latitude as input to avoid repeated 
        computations of trigonometric functions.
```

boule/ellipsoid.py
```diff
@@ -146,6 +146,32 @@ def gravity_pole(self):
         )
         return result
 
+    def _prime_vertical_radius(self, sinlat):
```
leouieda, line 149:
```suggestion
    def prime_vertical_radius(self, sinlat):
```

Might as well make this public if we’re planning to use it in Harmonica.

boule/ellipsoid.py
```diff
@@ -176,15 +202,14 @@ def geodetic_to_spherical(self, longitude, latitude, height):
 
         """
         latitude_rad = np.radians(latitude)
-        prime_vertical_radius = self.semimajor_axis / np.sqrt(
-            1 - self.first_eccentricity ** 2 * np.sin(latitude_rad) ** 2
-        )
+        coslat, sinlat = np.cos(latitude_rad), np.sin(latitude_rad)
+        prime_vertical_radius = self._prime_vertical_radius(sinlat)
```
leouieda, line 206:
```suggestion
        prime_vertical_radius = self.prime_vertical_radius(sinlat)
```

boule/ellipsoid.py
```diff
@@ -264,7 +289,7 @@ def normal_gravity(self, latitude, height):
             The (geodetic) latitude where the normal gravity will be computed
             (in degrees).
         height : float or array
-            The ellipsoidal (geometric) height of computation point (in meters).
+            The ellipsoidal (geometric) height of computation point (in meters)
```
leouieda, line 292: ?

boule/ellipsoid.py
santisoler: Nice catch! Seems I forgot to change them after drafting...

boule/ellipsoid.py
santisoler: Yes! I was thinking about that after I opened the PR, I completely agree.

boule/ellipsoid.py
santisoler: Please ignore... some unwanted vim keybinding maybe?

santisoler:
> I like the thought of just asking the user to do the sine calculation. 

:+1:

> We might as well make this a public method anyway. Left a few suggestions for the docstring.

Yes, I agree. I even though about making it public after I opened the PR yesterday.

> Still needs some tests. One way to test this is to calculate on the pole and equator and check against semi-major and semi-minor axis. Another is to try to find pre-computed values at mid latitudes for WGS84 or the like and test against that.

Yes, you're right about that. Even though the coordinate conversions tests are actually using it. we might want to add specific tests for this new method.
One little thing, the prime vertical radius at the poles is not equal to the semi-minor axis.
On Wikipedia you can find an alternative way of expressing the prime vertical radius:

equation

If latitude is 90 or -90, then the cosine is 0, therefore the prime vertical radius will be equal to:

image

That's why, when converting the coordinates to Cartesian, the z coordinate is also multiplied by (1 - e^2):

image

leouieda:
@santisoler looks good to go! I fixed the issues with the Mac builds. Merging and releasing 0.2.0

PR #37, opened 2020-07-10
Add a method for the ellipsoid geocentric radius

### LINKED ISSUE #34: Provide function that returns the radius as a function of latitude
It would be very useful to be able to compute the radius of the reference ellipsoid as a function of latitude (geocentric or geodetic). 

The use case I am thinking of is the following. Most spherical harmonic models of planetary topography provide the absolute radius of the surface. In many cases, it makes more sense to remove either a reference ellipse (which could be a sphere), or to remove the geoid. In pyshtools, I would like to create a grid of the radius of the reference ellipsoid like `grid = pysh.SHGrid.from_ellipsoid()`, and for this it would be most easy to input a boule ellipsoid and to use a boule function to compute the radius as function of latitude.

leouieda:
This is the distance from the center of the ellipsoid to its surface as
a function of geodetic latitude.

Fixes #34 

**Reminders**:

### DIFF
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -146,7 +146,7 @@ class Ellipsoid:
         )
         return result
 
-    def geocentric_radius(self, latitude):
+    def geocentric_radius(self, latitude, geodetic=True):
         r"""
         Distance from the center of the ellipsoid to its surface.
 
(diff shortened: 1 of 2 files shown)


leouieda:
@MarkWieczorek just to check, this is what you had in mind right?

MarkWieczorek:
That is about what I was looking for: But would it be possible to allow inputting geodetic or geocentric latitude? Perhaps you could just add a default parameter like `geodetic=True`. If False, the geocentric latitude would first be converted to geodetic.

I have almost never used geodetic coordinates, and when dealing with spherical harmonics, it is natural to use the gecentric latitude :)

leouieda:
Hum, I'll need to think about this a bit since we need the radius in the first place to convert geocentric to geodetic. So it doesn't seem to be straight forward. But there should be a way.

leouieda:
Alright, I worked out the equation and implemented the option for `geodetic=False` and some tests. I added a note saying that this is an exception since we generally favor solutions where the conversions are done by the user prior to input to simplify the code. But this wouldn't be an option here since there is no way to do the conversion without knowing the radius already.  

Seems like I got my math right :slightly_smiling_face:

PR #45, opened 2020-07-28
Add checks for ellipsoid parameters

### LINKED ISSUE #43: Check if Ellipsoid parameters are valid
**Description of the desired feature**

On #42 @leouieda proposed to add methods to the `Ellipsoid` class that checks whether the passed ellipsoid parameters are valid. This means that the values must fall under certain intervals to be able to create a valid ellipsoid, for example, the flattening should be a positive number.
Because these parameters are not defined through a regular constructor, like `__init__`, but rather defined using the `attrs` library, these methods should be decorated following [`attrs` documentation](https://www.attrs.org/en/stable/examples.html#validators).

We should check that (and raise errors if false):
- `flattening >= 0` and `flattening < 1`
- `semimajor_axis > 0`

We can also check that `geocentric_grav_const >= 0`, but for this I'll only raise a warning.

**Are you willing to help implement and maintain this feature?** Yes, but I can also help anyone that wants to implement it.

santisoler:
Add methods for validating ellipsoid parameters.
Raise error if flattening is negative or greater than 1.
Raise error if semimajor axis is not positive.
Raise warning if GM constant is negative.
Add tests functions for new features.

Fixes #43

**Reminders**:

### DIFF
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -1,6 +1,7 @@
 """
 Module for defining and setting the reference ellipsoid.
 """
+import warnings
 import attr
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


santisoler:
As discussed on #42 , we should decide if we should raise an error when passing a zero flattening or not.
We agreed that if flattening is tiny we should not raise an error, but a warning instead.

santisoler:
We have agreed to raise an error if `flattening == 0`, but only warn if `flattening < 1e-7`.

PR #48, opened 2020-07-28
Fix missing centrifugal term for sphere gravity at the equator

santisoler:
Fix the method that computes the norm of the gravity acceleration vector
on the equator of a sphere. The method used to return only the
gravitational attraction term, without the centrifugal acceleration
term. Add the additional term and update the test function, which now
compares gravity on poles and equator with the normal gravity formula.

**Reminders**:


santisoler:
Maybe we can extend the docstring of the `gravity_equator` method and explicitly say that they include both the gravitational attraction and the centrifugal acceleration. What do you think @leouieda ?

PR #51, opened 2020-07-29
Tests normal gravity against Somigliana equation

santisoler:
Add tests functions to compare ellipsoids' normal gravity on their surface against the Somigliana equation.

**Reminders**:

### DIFF
--- a/boule/tests/test_ellipsoid.py
+++ b/boule/tests/test_ellipsoid.py
@@ -8,6 +8,7 @@ import numpy as np
 import numpy.testing as npt
 
 from .. import Ellipsoid, ELLIPSOIDS
+from .utils import normal_gravity_surface
 
 
 ELLIPSOID_NAMES = [e.name for e in ELLIPSOIDS]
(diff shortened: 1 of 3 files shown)


leouieda:
@santisoler I made a few changes here: mostly moved the function to a `boule/tests/utils.py` module and made some adjustments to the function (calculating sin and cos only once).

leouieda:
We could remove the sphere test from this branch so we can merge it with tests passing. Then we can implement it in #52 so that we're confident that we can fix the sphere gravity. What do you think?

santisoler:
> We could remove the sphere test from this branch so we can merge it with tests passing. Then we can implement it in #52 so that we're confident that we can fix the sphere gravity. What do you think?

I agree. I will move the Somigliana test on the sphere to #52.

Thanks for the modifications, they are way better than what I drafted yesterday.

leouieda:
👍🏽 feel free to merge whenever you're ready

PR #57, opened 2020-10-09
Replace Travis and Azure for GitHub Actions

santisoler:
Add GitHub Actions workflows for checking style, testing and deploying
Boule, based on the workflows of Harmonica. Remove TravisCI and Azure
configuration files. Split make check target in two: `make black-check` and
`make flake8`. Ditch Travis and Azure badges on `README.rst`. Add new GitHub
Actions badge. Tests are run on every matrix configuration: all OSs and all
Python versions. Use pip to install dependencies instead of setting up miniconda.

Related to fatiando/contributing#19


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,175 @@
...
+        # If "optional", will install non-required dependencies in the build
+        # environment. Otherwise, only required dependencies are installed.
+        dependencies: [""]
+        include:
+          - os: ubuntu
+            python: 3.8
+          - os: macos
+            python: 3.8
+          - os: windows
+            python: 3.8
```
leouieda, line 49:
```suggestion
```

These seem redundant.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,175 @@
...
+      - name: Install the package
+        run: pip install --no-deps dist/*.whl
+
+      - name: Run the tests
+        run: |
+          make test
+
+      # -----------------------------------------------------------
+      # The following steps will run only on Ubuntu with Python 3.8
+      # -----------------------------------------------------------
```
leouieda, line 147: The docs build in 4 seconds so this is probably not necessary. The biggest time sink in the whole build is setting up miniconda actually. Switching to mamba or freezing dependencies is probably the most important thing we can do to speed these up.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,255 @@
...
+          user: __token__
+          password: ${{ secrets.PYPI_TOKEN}}
+
+  #############################################################################
+  # Publish the documentation to gh-pages
+  docs:
+    name: docs
+    runs-on: ubuntu-latest
+    # Only publish from the origin repository, not forks
+    if: github.repository == 'fatiando/harmonica'
```
leouieda, line 103:
```suggestion
    if: github.repository == 'fatiando/boule'
```

leouieda (commented):
Hi @santisoler thanks for doing this 🙂 I left a few comments. Feel free to merge once those are dealt with.

**Before merging**, we need to have the PyPI and Test PyPI tokens in the repository secrets.

leouieda:
> Before merging, we need to have the PyPI and Test PyPI tokens in the repository secrets.

I just added those. I also sent you an invitation for the Test PyPI accounts of all our packages.

leouieda:
@santisoler I tried using the Actions python with pip instead of conda and this cut down the test times from 3-4 minutes to ~1 min. It's really not necessary to use conda for Boule and many of our other packages that don't need complex dependencies. Rockhound might be the exception since now even numba can be pip installed. 

This might be a more significant speed up for CI than almost anything else we do.

santisoler:
Thanks @leouieda for the review and all the changes. I agree with the decision of install everything with `pip` instead of `conda`, specially for packages like Boule.

> Rockhound might be the exception since now even numba can be pip installed.

I don't know what you meant on this sentence. Did you mean Harmonica?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,149 @@
...
+          echo ""
+          echo "Generated files:"
+          ls -lh dist/
+
+      - name: Install the package
+        run: python -m pip install --no-deps dist/*.whl
+
+      - name: Run the tests
+        run: |
+          make test
```
santisoler, line 125:
```suggestion
        run: make test
```

leouieda:
> I don't know what you meant on this sentence. Did you mean Harmonica?

I mean that RockHound might require netCDF if we want to run any tests. But if not, then it should be fine. I think harmonica would be fine as well.

santisoler:
> I mean that RockHound might require netCDF if we want to run any tests. But if not, then it should be fine. I think harmonica would be fine as well.

Oh, now I get it! Sure, we can try it out on Harmonica and see if it works.

I think this is looking great. Please merge if no more changes should be made.

leouieda:
Fingers crossed that the deploy works 🤞🏽

santisoler:
Thanks for merging! Seems that deploys worked as charmed. Though one of the Mac OS tests was cancelled for no reason. Just rerun all jobs.

PR #58, opened 2020-10-14
Stop saving the docs HTML as build artifacts

leouieda:
It's a bit wasteful to store all this data that we will very likely
never use (downloading the cached page displaying them locally requires
a lot of clicking). So we might as well leave this out for now.

**Reminders**:


leouieda:
This is pretty minor so I'll merge it in so that we can focus on #59

PR #59, opened 2020-10-14
Use nox to manage development

leouieda:
[Nox](https://nox.thea.codes) is a tool that automates setting up virtual environments and running tasks inside them. It's a combination of `make` and virtual environments. Replace the machinery in our `Makefile` (which wouldn't work on Windows) with a cross-platform `noxfile.py`. Since nox handles the virtual environments, the `environment.yml` only needs to have python and nox. Plus, the environment creation code in the CI is no necessary anymore and we can run tests on multiple python versions locally as well.

**Reminders**:


leouieda:
I'm keeping the `Makefile` so that most instructions on the CONTRIBUTING.md still work but it's just a wrapper of the nox commands. We could remove it entirely and run `nox` instead.

leouieda:
Now that I think about it, I would probably organize the CI workflows differently:

* `style.yml`: stay the same
* `test.yml`: just have jobs for running the tests and submitting coverage. No docs builds here.
* `docs.yml`: build the docs and if on `master` or a release, deploy to gh-pages.
* `pypi.yml`: deploy to PyPI and TestPyPI.

This way we avoid having multiple duplicate builds of the docs. We could still build on all systems but I think it would be fine to only run it on ubuntu.

santisoler:
This looks great @leouieda! It's awesome to be capable of running tests on multiple Python versions locally without too much hassle.

For some reason my computer is skipping the py3.6 test:
```
nox > Session test-3.6 skipped: Python interpreter 3.6 not found.
```
Do you know why?

I did notice that even if you set up your conda environment, if you try to run `nox` without internet connection it fails because cannot install the dependencies. Is there a way to preinstall everything that `nox` needs beforehand so we don't depend on an internet connection for running the tests?

santisoler:
> I did notice that even if you set up your conda environment, if you try to run nox without internet connection it fails because cannot install the dependencies. Is there a way to preinstall everything that nox needs beforehand so we don't depend on an internet connection for running the tests?

Actually there is:

```
nox --install-only
```
https://nox.thea.codes/en/stable/usage.html?highlight=offline#skipping-everything-but-install-commands

noxfile.py
```diff
@@ -0,0 +1,212 @@
...
+    Build source and wheel packages for the project and install them
+    """
+    if Path("dist").exists():
+        shutil.rmtree("dist")
+    session.install("wheel")
+    session.run("python", "setup.py", "sdist", "bdist_wheel", silent=True)
+    wheel = [str(p) for p in Path("dist").glob("*.whl")]
+    assert len(wheel) == 1, f"More than 1 wheel present: {wheel}"
+    session.install("--no-deps", wheel[0])
+    # session.install("--no-deps", "-e", ".")
```
santisoler, line 95:
```suggestion
```

noxfile.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    Build source and wheel packages for the project and install them
+    """
+    if Path("dist").exists():
+        shutil.rmtree("dist")
+    session.install("wheel")
+    session.run("python", "setup.py", "sdist", "bdist_wheel", silent=True)
+    wheel = [str(p) for p in Path("dist").glob("*.whl")]
+    assert len(wheel) == 1, f"More than 1 wheel present: {wheel}"
+    session.install("--no-deps", wheel[0])
```
santisoler, line 94: I'm getting an error from this assertion when running `nox --install-only` because the `wheel` is an empty list.

```suggestion
    if wheel:
        assert len(wheel) == 1, f"More than 1 wheel present: {wheel}"
        session.install("--no-deps", wheel[0])
```

santisoler (changes requested):
I left some suggestions down here. I'm not entirely convinced about `nox`, but I like the idea! I think I just need to spend some time with it to get used to it.

.github/workflows/continuous-integration.yml
```diff
@@ -39,13 +39,7 @@ jobs:
       matrix:
         os: [ubuntu, macos, windows]
         python: [3.6, 3.7, 3.8]
```
santisoler, line 41: Taking into account that `nox` is capable of running the tests on every Python version we want, do we really need to set Python versions on the `matrix`? Maybe we can save spare some CI time by reducing the number of virtual machines that we build on every run.

leouieda: I thought we might but this is related to why it's skipping your 3.6 build. Nox doesn't download new Python versions unless using `conda`. So it will only use what's in the system already. And I don't think we can set multiple Python's on Actions. So it's better to keep it this way. But the time saved by not using miniconda for everything is much larger than spinning up a vm.

leouieda:
About running offline:

* The `--install-only` lets you build the environments only without running tests. 
* I'm hacking on this in Erizo and will copy over some improvements to allow skipping install steps so that you can work offline

The default is making fresh environments but we can reverse that by setting `nox.options.reuse_existing_virtualenvs=True`. We can always disable on the command line with `--no-reuse-existing-virtualenvs`. I'll set this on and then we can skip the `-r` without much fear of nuking the environments.

leouieda:
Alright, with the new file after 08e23c4 you can now:

* Reuse environments by default
* Pass in the `skip-install` option to any session to skip the install step
* Run all checks or selected checks by passing arguments to the `check` session.

Need to add these instructions to the docstring still.

.github/workflows/continuous-integration.yml
santisoler: You convinced me! Let's keep it this way for now then!

leouieda:
@santisoler if you could give this a try locally that would be very helpful. See if you like it and if there are any other pain points we can try to resolve

noxfile.py
```diff
@@ -0,0 +1,263 @@
...
+    )
+
+
+@nox.session()
+def build(session):
+    """
+    Build source and wheel distributions for the project
+    """
+    install_requirements(session, ["build"])
+    packages = build_project(session, install=False)
```
santisoler, line 143: `packages` is assigned but never used. Don't know if you forgot to use after this line or if we can ditch the `packages` variable.

noxfile.py
```diff
@@ -0,0 +1,263 @@
...
+    Install dependencies from the requirements files specified in REQUIREMENTS.
+
+    *requirements* should be a list of keywords defined in the REQUIREMENTS
+    dictionary. Set *package_manager* to "conda" if using the conda backend for
+    virtual environments.
+    """
+    if package_manager not in {"pip", "conda"}:
+        raise ValueError(f"Invalid package manager '{package_manager}'")
+    if session.posargs and "skip-install" in session.posargs:
+        session.log(f"Skipping install steps.")
```
santisoler, line 194: We don't need to use f-string here.
```suggestion
        session.log("Skipping install steps.")
```

noxfile.py
```diff
@@ -0,0 +1,263 @@
...
+    Build source and wheel packages for the project and returns their path.
+    If 'install==True', will also install the package.
+    """
+    session.log("Build source and wheel distributions:")
+    if Path("dist").exists():
+        shutil.rmtree("dist")
+    session.run("python", "setup.py", "sdist", "bdist_wheel", silent=True)
+    packages = list(Path("dist").glob("*"))
+    if install:
+        session.install("--force-reinstall", "--no-deps", str(packages[0]))
```
santisoler, line 254: I'm getting an error when running `nox --install-only` because `packages` is empty.
Maybe we can check if `packages` is not empty before installing the package.
```suggestion
    if install and packages:
        session.install("--force-reinstall", "--no-deps", str(packages[0]))
```

santisoler (changes requested):
This is looking great @leouieda!

I'm linking the simplicity of nox (and writing workflows in Python is much easier for me than writing Makefiles). I added some suggestions, please see if you like them.

I would like to add a few `make` targets:
- `make nox-install`: for creating the virtual environments and installing dependencies (leaving everything ready for offline runs).
- `make nox-clean`: for cleaning the `.nox` directory.
- `make nox-update`: for cleaning and then installing the environments. I think this might be useful not only for offline runs but also for updating any old-dated nox environments.

Also, I would like to add a `serve` session for serving the built docs locally.

Let me know what do you think.

noxfile.py
```diff
@@ -0,0 +1,263 @@
...
+"""
+from pathlib import Path
+import shutil
+
+import nox
+
+
+# CONFIGURE NOX
+# Specify session that will execute when running with no arguments
+nox.options.sessions = ("check", "test")
```
santisoler, line 16: Let `nox` to run every session by default.
```suggestion
```

leouieda:
@santisoler I made a few changes:

* Added instructions to the noxfile.py docstring
* Added a `show` option to the docs build that opens the HTML in a web browser. I opted for this instead of serving since it would block the process and the noxfile would just hang. This works just as well and is easily cross-platform using `import webbrowser`.
* Added rules to the Makefile for `nox-clean` and `nox-install`. The update command is not needed since it would be a clean and install anyway. 

What do you think?

santisoler (approved):

santisoler:
This is looking great!

>     * Added instructions to the noxfile.py docstring

I like that we can list all the available targets with `nox -l`. We definitely should add this to the Contributing instructions once we ditch the `make` instructions.
 
>     * Added a `show` option to the docs build that opens the HTML in a web browser. I opted for this instead of serving since it would block the process and the noxfile would just hang. This works just as well and is easily cross-platform using `import webbrowser`.

Much better than before!

>     * Added rules to the Makefile for `nox-clean` and `nox-install`. The update command is not needed since it would be a clean and install anyway.

Sure, I was thinking to add the `nox-update` target calling the `nox-clean` and then `nox-install`, but we don't actually need another target to do so, we can use `make nox-clean nox-update` on a single line.

I think this is ready to go! Nice work @leouieda !

leouieda:
Thanks @santisoler! Merging this in. I'll work on a follow up to organize the CIs a bit. Only thing I changed is that `setuptools_scm` doesn't need to be a run time dependency.

PR #61, opened 2020-10-26
Replace versioneer with setuptools-scm

santisoler:
Remove all versioneer related files. Update setup.py to work with
setuptools-scm. Update boule/version.py to work with setuptools-scm.
Remove setup.cfg and move all flake8 configuration to .flake8 file. Add
setuptools-scm to requirements.txt and environment.yml.

Related to fatiando/maintenance#4


leouieda:
👍🏽 thanks, Santiago! Shall we merge this first and then #59?

santisoler:
Sure! I'll merge this right away!

PR #65, opened 2020-10-27
Refactor workflows to separate building the docs

leouieda:
Instead of running the docs build every time, leave it to one workflow
which can then reuse the HTML artifacts to deploy. Separate the PyPI
deploy in a separate workflow as well.

**Reminders**:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,14 +1,16 @@
-# Configuration for publishing archives to PyPI and documentation pages to
-# GitHub Pages using GitHub Actions.
+# Build the documentation and deploy to GitHub Pages using GitHub Actions.
 #
 # NOTE: Pin actions to a specific commit to avoid having the authentication
 # token stolen if the Action is compromised. See the comments and links here:
 # https://github.com/pypa/gh-action-pypi-publish/issues/27
 #
-name: deploy
+name: docs
 
-# Only run for pushes to the master branch and releases.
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -40,14 +37,14 @@ jobs:
         with:
           # Need to fetch more than the last commit so that versioneer can
```
santisoler, line 38:
```suggestion
          # Need to fetch more than the last commit so that setuptools-scm can
```

.github/workflows/docs.yml
```diff
@@ -40,14 +37,14 @@ jobs:
...
           # create the correct version string. If the number of commits since
-          # the last release is greater than this, the version will still be wrong.
+          # the last release is greater than this, the version still be wrong.
```
santisoler, line 40: "the version still be wrong" sounds weird, shouldn't the _to be_ verb be conjugated?

santisoler: On `pypi.yml` the sentence is _"the version will still be wrong"_.

santisoler (approved):
Nice clean up of the Actions! They look much better now!

Left a few comments below.

.github/workflows/docs.yml
leouieda: Yeah, this was a git merge error 🙁

leouieda, line 40:
```suggestion
          # the last release is greater than this, the version will still be wrong.
```

PR #67, opened 2020-11-02
Add license and copyright notice to every .py file

santisoler:
Related to fatiando/maintenance#10

**Reminders**:

### DIFF
--- a/boule/__init__.py
+++ b/boule/__init__.py
@@ -1,9 +1,3 @@
-# Copyright (c) YEAR The PROJECT Developers.
-# Distributed under the terms of the BSD 3-Clause License.
-# SPDX-License-Identifier: BSD-3-Clause
-#
-# This code is part of the Fatiando a Terra project (https://www.fatiando.org)
-#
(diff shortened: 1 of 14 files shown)


boule/__init__.py
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2019 The Boule Developers.
+# Distributed under the terms of the BSD 3-Clause License.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# This code is part of the Fatiando a Terra project (https://www.fatiando.org)
```
leouieda, line 5: Should we add an empty comment line here to separate this from the pylint configuration?

```suggestion
# This code is part of the Fatiando a Terra project (https://www.fatiando.org)
#
```

leouieda: Or maybe remove the empty comment line before this. Either one would do.

santisoler: You are right. I like the empty line between the license notice and the `This code is part...` line. So we should add an empty comment line after that one.

Issue #46, opened 2020-07-28
Option to return gravity in SI units


leouieda:
**Description of the desired feature**

In #42, a point was raised that it would be convenient to have an option the gravity calculation methods return values in SI units. This would be a temporary fix until we have proper unit support (#23) but it would be useful to have. We can always remove/deprecate it once units are added. 

It would mean adding the argument `si_units=False` to the `normal_gravity` method of `Ellipsoid` and `Sphere`. 

**Are you willing to help implement and maintain this feature?** Yes but welcome anyone who would like to do it

PR #69, opened 2020-11-09
Add option to return normal gravity in SI units

dabiged:
This PR adds an option to return gamma in SI units for the normal_gravity method in both the ellipsoid and sphere classes as per #46 .
I have included an example in the sphere docstring and added tests for both classes.

**Reminders**:

- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [x] Add new public functions/methods/classes to `doc/api/index.rst` and the base `__init__.py` file for the package.
- [x] Write detailed docstrings for all functions/classes/methods. It often helps to design better code if you write the docstrings first.
- [x] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.

### DIFF
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -436,7 +436,7 @@ class Ellipsoid:
         height : float or array
             The ellipsoidal (geometric) height of computation the point (in
             meters).
-        si_units : bool 
+        si_units : bool
             Return the value in mGal (False, default) or SI units (True)
 
         Returns
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -419,7 +419,7 @@ class Ellipsoid:
         big_d = k * radius * cos_latitude / (k + self.first_eccentricity ** 2)
         return k, big_z, big_d
 
-    def normal_gravity(self, latitude, height):
+    def normal_gravity(self, latitude, height, si_units=False):
         """
         Calculate normal gravity at any latitude and height.
 
@@ -436,6 +436,8 @@ class Ellipsoid:
         height : float or array
             The ellipsoidal (geometric) height of computation the point (in
             meters).
+        si_units : bool 
+            Return the value in mGal (False, default) or SI units (True)
 
         Returns
         -------
@@ -460,6 +462,8 @@ class Ellipsoid:
         )
         term3 = -cosbeta_l2 * b_l * self.angular_velocity ** 2
         gamma = (term1 + term2 + term3) / big_w
+        if si_units:
+            return gamma
         # Convert gamma from SI to mGal
         return gamma * 1e5
 
--- a/boule/ellipsoid.py
+++ b/boule/ellipsoid.py
@@ -419,7 +419,9 @@ class Ellipsoid:
         big_d = k * radius * cos_latitude / (k + self.first_eccentricity ** 2)
         return k, big_z, big_d
 
-    def normal_gravity(self, latitude, height, si_units=False):
+    def normal_gravity(
+        self, latitude, height, si_units=False
+    ):  # pylint: disable=too-many-locals
         """
         Calculate normal gravity at any latitude and height.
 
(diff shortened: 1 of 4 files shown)


boule/tests/test_ellipsoid.py
```diff
@@ -202,6 +202,20 @@ def test_normal_gravity_arrays(ellipsoid):
...
+    heights = np.zeros(3)
+    latitudes = np.array([-90, 90, 0])
+    gammas = np.array(
+        [ellipsoid.gravity_pole, ellipsoid.gravity_pole, ellipsoid.gravity_equator]
+    )
+    npt.assert_allclose(
+        gammas, ellipsoid.normal_gravity(latitudes, heights, si_units=True), rtol=rtol
+    )
+
+
```
leouieda, line 218: This could be replaced by another `parametrize` in line 190:

```python
@pytest.mark.parametrize("si_units", [False, True], ids=["mGal", "SI"])
@pytest.mark.parametrize("ellipsoid", ELLIPSOIDS, ids=ELLIPSOID_NAMES)
def test_normal_gravity_arrays(ellipsoid, si_units):
    "Compare normal gravity passing arrays as arguments instead of floats"
    ...
```

You'd also need an `if` to check if you need to convert the `gammas` list.

boule/tests/test_sphere.py
```diff
@@ -153,6 +153,26 @@ def test_normal_gravity_pole_equator(sphere):
...
+        gamma_pole, sphere.normal_gravity(-90, height, si_units=True), rtol=rtol
+    )
+    npt.assert_allclose(
+        gamma_pole, sphere.normal_gravity(90, height, si_units=True), rtol=rtol
+    )
+    npt.assert_allclose(
+        gamma_eq, sphere.normal_gravity(0, height, si_units=True), rtol=rtol
+    )
+
+
```
leouieda, line 175: Same thing here:

```python
@pytest.mark.parametrize("si_units", [False, True], ids=["mGal", "SI"])
def test_normal_gravity_pole_equator(sphere, si_units):
    """
    Compare normal gravity values at pole and equator
    """
```

leouieda (changes requested):
@dabiged looking good! I like that you already added the docstring sections and some doctests. I have a few suggestions for the tests. Basically, these can be combined with previous tests instead of having to copy the test function (using `pytest.mark.parametrize`).

dabiged:
First time using pytest :blush: 
Thanks for the pointers. It is much cleaner now.

boule/ellipsoid.py
```diff
@@ -419,7 +419,7 @@ def _spherical_to_geodetic_terms(self, spherical_latitude, radius):
         big_d = k * radius * cos_latitude / (k + self.first_eccentricity ** 2)
         return k, big_z, big_d
 
-    def normal_gravity(self, latitude, height):
+    def normal_gravity(self, latitude, height, si_units=False):
```
leouieda, line 422:
```suggestion
    def normal_gravity(self, latitude, height, si_units=False): # pylint: disable=too-many-locals
```

leouieda: This should make pylint stop complaining about too many variables. In this case, there is no way around it. You might need to run black again after this if the line is too long.

leouieda:
> First time using pytest 

@dabiged it's pretty sweet, right? I only learned about these tricks several years after picking it up.

All that is left now is pleasing pylint. I left a comment above and that should do the trick.

Also, please feel free to add yourself to `AUTHORS.md` (if you'd like of course). We use that to give attribution when we publish Zenodo archives of new versions. If you add yourself, we'd be happy to list you as co-author on the archives 🙂 This is entirely optional, of course.

leouieda (approved):

leouieda:
👍🏽 merged! Loving all the PRs coming in @dabiged 🙂

Issue #47, opened 2020-07-28
Implement TriAxialEllipsoid class


MarkWieczorek:
Boule currently supports biaxial ellipsoids (under the classname `Ellipsoid`) and spheres (which are a subclass of `Ellipsoid`). However, one major omission, which is important for all of the planetary moons, is the case of triaxial ellipsoids.

Though it may be difficult to implement all the features of the present ellipsoid class, there are several easy functions that would be of immediate use:

1. `radius` of the body as a function of (geocentric) latitude and longitude,
2. Ellipsoid parameters `omega`, `GM`, `a`, `b`, and `c`,
3. Some easily derived quantities, such as ellipticities, flattenings, etc.,
4. Orbital parameters of the body, such as the eccentricity, inclination, and semimajor axis,
5. The `volume` of the body,
6. The tidal potential raised by the primary (assuming the moon is in synchronous rotation), and the centrifugal potential (non-zero inclinations and eccentricies would be useful as well, but their computation are non-trivial).

Other quantities that geodecists use on Earth (like normal gravity, geodetic coordinates, etc), may take significant effort to implement, but would be of use for only a handful of experts (who might decide to implement these themselves).

At this point, given that Boule would have three types of ellipsoids, it might be useful to make some kind of factory function called `Ellipsoid` that would automatically instantiate one of the three classes based on the input parameters:
* Sphere
* BiAxialEllipsoid
* TriAxialEllipsoid

leouieda:
@MarkWieczorek we could add the class with only the geometric properties and raise `NotImplementedError` on the gravity and coordinate conversion calculations. Since use of this is a bit niche then it wouldn't be a major problem if we start without having these methods. Do you have any reference for these calculations? Having it linked here would help anyone who is picking this up.

I hadn't thought about the orbits. This might be better done in stages, with the geometric parts first and then adding orbital parameters and calculations later on. That falls way outside of my expertise. Would this be added to the biaxial ellipsoid and sphere as well?

We can tackled the factory function after the class is available and we can see what limitations might come up.

MarkWieczorek:
The basic geometric properties are found here:

https://en.wikipedia.org/wiki/Ellipsoid

The article notes that

> For geodesy, geodetic latitude, the angle between the vertical and the equatorial plane, is most commonly used. Geodetic latitude is not defined for a general ellipsoid because it depends upon longitude.

So this simplifies our lives.

As for the orbital parameters: The initial step would be just to add them to a dictionary or as an attribute.

[...]

PR #72, opened 2020-11-10
Add TriaxialEllipsoid class with geometric parameters

dabiged:
Skeleton class for a Triaxial Ellipsoid. I have implemented the basic framework and a number of properties from Issue #47 . 
More advanced methods raise a NotImplementedError. 
I have taken formula from wikipedia and this paper: https://www.sciencedirect.com/science/article/abs/pii/S0019103584711183?via%3Dihub

ToDo:
 - [x] Internal Doctests using a generic watermelon triaxial ellipsoid.
 - [x] Test within the class that semimajor > semimedium > semiminor.
 - [x] Tests to the test directory.
 - [x] realization of a triaxial ellipsoid (Vesta / Eros?)
 - [ ] Examples to the gallery

Submitting this Pull request early to get feedback. 

- Should the class be called Tri**A**xialEllipsoid or Tri**a**xialEllipsoid
- Is the current watermelon example suitable?

**Reminders**:


dabiged:
I would love some feedback on this before I go much further.
Implementing things like a radius method, or anything more advanced seems to require 
elliptic integrals which I do not feel comfortable doing.

boule/triaxialellipsoid.py
```diff
@@ -0,0 +1,254 @@
...
+    def volume(self):
+        """
+        The volume of a triaxial ellipsoid :math: `V = /frac{4}{3} pi a b c ` [meters^3]
+        """
+        return (
+            4
+            * np.pi
+            / 3
+            * (self.semimajor_axis * self.semimedium_axis * self.semiminor_axis)
+        )
```
santisoler, line 213: Nice! If we add this method on `TriaxialEllipsoid` we should do the same on `Ellipsoid` and `Sphere`, but that's for a separate PR.

boule/triaxialellipsoid.py
```diff
@@ -0,0 +1,254 @@
...
+    @property
+    def gravity_pole(self):
+        "The norm of the gravity vector on the ellipsoid at the poles [m/s²]"
+        raise NotImplementedError
+
+    def radius(self, latitude, longitude):
+        r"""
+        Return the radius of the Triaxial Ellipsoid at the given latitude and longitude [metres]
+        """
+        raise NotImplementedError
```
santisoler, line 224: We don't need to define a `radius` for the `TriaxialEllipsoid`. The `radius` is only suited for the `Sphere`.
```suggestion
```

santisoler (changes requested):
Hi @dabiged! I'm really happy to see an open PR for Triaxial Ellipsoids.

I don't have too much knowledge on this type of ellipsoids, maybe @MarkWieczorek can give a higher quality review. Nevertheless I'll leave some inline comments (probably more related to code than to the underlying math).

santisoler:
I'm wondering if we can benefit from creating some inheritance, I mean make `Ellipsoid` a subclass of `TriaxialEllipsoid`. Even though from a mathematic perspective is true (any biaxial ellipsoid is a triaxial ellipsoid with one pair of axis of equal lenght), maybe it could not bring any advantage from a code design perspective.

cc @leouieda

boule/triaxialellipsoid.py
dabiged: I have added this as issue #75 so I don't forget!

boule/triaxialellipsoid.py
dabiged: This method was mentioned in the feature request. 

I assumed this was a vector running from the centre of the ellipsoid (aka the origin [is it called the geocentric location??]) to a point on the surface at a given latitude and longitude. 
For a sphere this is a constant.
For a biaxial ellipsoid this maps to an ellipse along lines of longitude and a constant on lines of latitude.
For a triaxial ellipsoid I think it is this value (source: wikipedia page on ellipsoids).

image

Should this method be called something else?

boule/triaxialellipsoid.py
```diff
@@ -0,0 +1,254 @@
...
+        Check if semimajor axis is the largest of the 3 axes.
+        """
+        if self.semiminor_axis > value:
+            raise ValueError(
+                f"Invalid semi-minor / semi-major axis combination.                The semimajor axis must be larger than the semi-medium axis.                Semi-major axis was '{value}' and the semi-minor axis was '{self.semiminor_axis}'"
+            )
+        if self.semimedium_axis > value:
+            raise ValueError(
+                f"Invalid semi-medium / semi-major axis combination.                 The semimajor axis must be larger than the semi-medium axis.                 Semi-major axis was '{value}' and the semi-medium axis was '{self.semimedium_axis}'"
+            )
```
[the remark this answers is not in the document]
dabiged: This is my first time using the `attr` toolset. I have no idea how to write this at the attr documentation seems quite sparse. Can you give me any idea where to start?

leouieda:
@santisoler @dabiged this is bringing up some interesting flaws in our current design. I’m usually hesitant to add a lot of inheritance. It’s rarely a good solution in my experience. Instead, see #76 for a proposal to have a single class. I think this would fix a lot of issues with methods taking unused arguments.

boule/triaxialellipsoid.py
leouieda: This is akin to the `geocentric_radius’s of `Ellipsoid` https://www.fatiando.org/boule/dev/api/generated/boule.Ellipsoid.html#boule.Ellipsoid.geocentric_radius

santisoler: Sorry for the misunderstanding: I though you were trying to implement a not implemented method in replacement of the `radius` attribute of the `Sphere`.
Now that I know what you meant by `radius`, I completely agree that the class should have this method.
Nevertheless, `radius` is a very misleading name. The `Ellipsoid` class has a [`geocentric_radius`](https://www.fatiando.org/boule/latest/api/generated/boule.Ellipsoid.html#boule.Ellipsoid.geocentric_radius) method which returns the quantity you describe. That name explicitly says that we are about to return the norm of the _radius_ vector to the surface of the ellipsoid measured from a geocentric coordinate system, so the origin is at the center of the ellipsoid.

In the `TriaxialEllipsoid`, the `geocentric_radius` should get the `longitude` and the `latitude`, as you already did. But, remember that on every Fatiando library we keep the following order for coordinates: `longitude`, `latitude`, `height`, ... or `easting`, `northing`, `height`, ..., except of very special cases where the reason is well documented.

So feel free to ignore this suggestion and please implement the `geocentric_radius` as you described.

MarkWieczorek:
I am ok with `geocentric_radius`, but at the same time, in spherical coordinates (r, theta, phi), r is "radius". Maybe geodesists aren't used to working with spherical coordinates, so having this be explicit is ok.

boule/triaxialellipsoid.py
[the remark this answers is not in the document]
santisoler: We could add a new `semimedium_axis` validator that checks the `semimedium_axis` against the other two. Something like:
```python
    @semimedium_axis.validator
    def _check_valid_axes(self, semimedium_axis, value):
        # Check if semimajor axis is the largest of the 3 axes.
        if self.semiminor_axis > value or value > self.semimajor_axis:
            raise ValueError("Invalid set of axes: ....")
```
This way we can have a single `ValueError` for any possible combination of invalid set of axes.
Although it's weird to have a `semimedium_axis.validator` decorator for a method that checks the entire set of axes, I couldn't came out with a better solution.

cc @leouieda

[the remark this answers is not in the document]
santisoler: We could also add another validator that raise warning if `semimedium_axis` is equal to `semiminor_axis` or equal to `semimajor_axis` and instruct people to use the `Ellipsoid` class instead.

boule/triaxialellipsoid.py
dabiged: Thanks for clearing up the misunderstanding there.
I was completely wrong and didn't see geocentric_radius method.

I have removed the radius method from the latest commit.

leouieda:
Alright, I finished the few things left over from this PR. Removed all of the not implemented methods instead to simplify things. We can add them later when we have the need for such calculations.

PR #73, opened 2020-11-11
Add normal gravitation method to Sphere

### LINKED ISSUE #54: Add `normal_gravitation` method to `Sphere`
**Description of the desired feature**

After #52 we decided that the `Sphere.normal_gravity` method should return the norm of the gradient of the gravity potential (gravitational + centrifugal). Nevertheless, as @MarkWieczorek pointed out, we might also want a `normal_gravitation` method that returns the norm of the gradient of the gravitational potential (without the centrifugal term), a.k.a the norm of the gravitational acceleration.

The norm of the gravitational acceleration of a sphere can easily be computed as:

image

where `r` is the distance between the observation point and the center of the sphere.

The new method should be able to return the norm of the gravitational acceleration for any given latitude and height (even though the gravitational acceleration of the sphere does not depend on latitude, we might want to keep the same notation for a future `Ellipsoid.normal_gravitation`):

```python
def normal_gravitation(self, latitude, height):
    ...
    return normal_gravitation
```

This issue is inspired on #19 , which may require some more work and discussion, so it's better to start with this part which seems easier.

Yes, I'm willing to help anyone that wants to implement it (yes, I'm talking to you)! :rocket

dabiged:
This includes the `si_units` flag as discusses in #69.
Fixes #54

**Reminders**:

- [x] Write detailed docstrings for all functions/classes/methods. It often helps to design better code if you write the docstrings first.
- [x] If adding new functionality, add an example to the docstring, gallery, and/or tutorials.


boule/sphere.py
```diff
@@ -208,6 +208,44 @@ def normal_gravity(self, latitude, height, si_units=False):
         # Convert gamma from SI to mGal
         return gamma * 1e5
 
+    def normal_gravitation(self, height, latitude=0, si_units=False):
+        r"""
+        The norm of the gravitational acceleration.
```
leouieda, line 213:
```suggestion
        Calculate the norm of the gravitational acceleration of the sphere.
```

boule/sphere.py
```diff
@@ -208,6 +208,44 @@ def normal_gravity(self, latitude, height, si_units=False):
         # Convert gamma from SI to mGal
         return gamma * 1e5
 
+    def normal_gravitation(self, height, latitude=0, si_units=False):
```
leouieda, line 211:
```suggestion
    def normal_gravitation(self, latitude, height, si_units=False):
```

leouieda: This should be almost identical to `normal_gravity` so that it's easier for users to switch between them if needed.

boule/sphere.py
```diff
@@ -208,6 +208,44 @@ def normal_gravity(self, latitude, height, si_units=False):
...
+        r"""
+        The norm of the gravitational acceleration.
+
+        Computes the magnitude of the gravitational acceleration generated by
+        the sphere at the given latitude :math:`\theta` and height :math:`h`:
+
+        .. math::
+
+            \gamma(\theta, g) =
```
leouieda, line 220:
```suggestion
            \gamma(\theta, h) =
```

boule/sphere.py
```diff
@@ -208,6 +208,44 @@ def normal_gravity(self, latitude, height, si_units=False):
...
+            \frac{GM}{(R + h)^2} \right
+
+        in which :math:`R` is the sphere radius, :math:`G` is the gravitational
```
leouieda, line 223:
```suggestion
        in which :math:`R` is the sphere radius, :math:`G` is the gravitational
        constant, and :math:`M` is the mass of the sphere.
```

boule/sphere.py
```diff
@@ -208,6 +208,44 @@ def normal_gravity(self, latitude, height, si_units=False):
...
+
+        Parameters
+        ----------
+        height : float or array
+            The height (above the surface of the sphere) of the computation
+            point (in meters).
+        latitude : float or array
+            The latitude where the normal gravity will be computed (in
+            degrees). This value is not used in the computation but is kept
+            here for backward compatibility.
```
leouieda, line 233: Need to be swapped.

leouieda, line 233:
```suggestion
            degrees). This value is not used in the computation but is kept
            here for compatibility with ``normal_gravity``. Can be ``None``.
```

boule/sphere.py
```diff
@@ -208,6 +208,44 @@ def normal_gravity(self, latitude, height, si_units=False):
...
+        si_units : bool
+            Return the value in mGal (False, default) or SI units (True)
+
+        Returns
+        -------
+        gamma : float or array
+            The normal gravity in mGal.
```
leouieda, line 240:
```suggestion
            The normal gravitation in mGal.
```

doc/ellipsoids/moon.rst
```diff
@@ -22,3 +22,6 @@ The parameters of the Lunar spheroid were obtained from [Wieczorek2015]_:
     >>> # Angular velocity
     >>> print("{:.7e}".format(MOON.angular_velocity))
     2.6617073e-06
+    >>> # Acceleration due to gravity alone
+    >>> print("{:.3f}".format(MOON.normal_gravitation(height=0, si_units=True)))
+    1.625
```
leouieda, line 27: This would be better in the docstring of `Sphere` where we already have an example of normal gravity.

leouieda (changes requested):
@dabiged thanks for yet another PR 🚀 

Left a few comments to improve compatibility with `normal_gravity` and with future `normal_gravitation` for the ellipsoid. The idea is that we can pass either a `Sphere` or `Ellipsoid` to someone and most methods should work exactly the same (same name, same arguments).

leouieda:
I edited your PR description a bit. If you use the keyword `Fixes`, GitHub will automatically close the issue when this PR is merged.

boule/sphere.py
dabiged: If I understand this correctly you want the order of these two parameters the same, which I can get onboard with.

I would like to default the latitude to a value so that it is not required. It isn't used in the calculation at all, and may confuse users if it asks for a non-relevant parameter.

leouieda: So this is a bit of a conundrum:

1. For latitude to be optional, it has to come after `height` but that means that `Sphere.normal_gravitation` will be different from `Sphere.normal_gravity`, `Ellipsoid.normal_gravity`, and `Ellipsoid.normal_gravitation` (which is not implemented yet). 
2. For `Sphere.normal_gravitation` to be compatible with everything else, we need to require latitude even if we're not using it (but it could be set to `None` by the user).

Unfortunately, there is no good way to resolve both points simultaneously. In fact, this will be even worse when `TriaxialEllipsoid` has gravity calculations because it will require `longitude` as well. 

To me, 2. is the better option because if `Sphere` and `Ellipsoid` behave exactly the same, then using `boule.MOON` or `boule.WGS84` doesn't matter for user scripts or other libraries using Boule. So I could have a gravity processing function that takes a Boule realization and give it `MOON` with lunar data or `WGS84` with Earth data without the function needing to know the difference. We already do this with the coordinate conversions that don't touch `longitude` but require it anyway for compatibility with other libraries.

Initially, we were going to have everything be the `Ellipsoid` class. We only introduced `Sphere` because there are singularities in the `normal_gravity` of ellipsoids with zero flattening.

dabiged: The latest commit makes your recommended change.

Is the plan to add longitude to both the Sphere and Ellipsoid classes once we have a functioning triaxial Ellipsoid?

dabiged:
Can someone explain what the automated check here is complaining about? I don't know how to fix this.

boule/sphere.py
leouieda: Thanks, @dabiged. I hope this makes sense but I'm open to suggestions on alternative ways to resolve this.

> Is the plan to add longitude to both the Sphere and Ellipsoid classes once we have a functioning triaxial Ellipsoid?

I only realized this as I was writing the reply above so I think we might have to do that and break the API a bit. We can wait until TriaxialEllipsoid has gravity calculations, though.

leouieda:
> Can someone explain what the automated check here is complaining about? I don't know how to fix this.

It's complaining that only 66% percent of your changes are executed during testing. This means that some parts of the code are never executed by the tests. Run the tests locally and you will see exactly which lines don't get executed as part of the test results. This is probably the if statement for SI units. Take a look at `boule/tests` for examples of test functions.

boule/sphere.py
santisoler: This is tricky. I like to keep the same method design on every ellipsoid class, so if I have a workflow for something on the Earth I can simply change it to Mars without getting weird errors because I'm passing an extra `latitude` argument to `normal_gravity`. The same for any triaxial ellipsoid complaining about a missing `longitude` argument.

But I know it's weird to have methods with arguments that they don't actually use. This may be not only confusing when using it, but could also create the misleading sensation that the normal gravity of the Moon varies on `latitude`.

What if we keep the best of both worlds by making our methods sufficiently smart through variable-length arguments:
```python
def normal_gravity(*coordinates, si_units=False):
    """
    """
    ...
```
This way the methods could take any number of coordinates (up to three, obviously) be sufficiently smart to know what they mean.
- If `len(coordinates) == 1` -> `height`
- If `len(coordinates) == 2` -> `latitude` and `height`
- If `len(coordinates) == 3` -> `longitude`, `latitude` and `height`

With this solution we don't need to break the API.
For obvious reasons, an `Ellipsoid` could only get 2 or 3 coordinates and should raise an error if `len(coordinates) == 1`.
The same for `TriaxialEllipsoid`: it should only take 3 coordinates, and in that case we can just use the `normal_gravity(longitude, latitude, height, ...)` syntax.

I'm brainstorming, so feel free to disagree with this idea or, even better, improve it.

dabiged: I have had a go at implementing something like this in (https://github.com/fatiando/boule/pull/73/commits/c0104c1db265783b68c01de9dd3d1bb3b89b4463.) I think we need to use **kwargs to capture the arguments as keywords as a `*coordinate` style does not specify which coordinate is which height/latittude/longitude. I think 
`  
def normal_gravity(longitude=12, latitude=24, height=36 )
`
is far cleaner and less ambiguous than:
`    def normal_gravity([12, 24, 36])`

I have also updated the docstring in the sphere class to show that you can now pass lat/long/height to normal_gravity, or just height. 

There needs to be more work on this to deal with weird arguments etc, but is this along the lines of what you were thinking @santisoler ?

leouieda: The trade-off there is that with this type of code tab-completion no longer works and users have to read the entire docstring to figure out the order of arguments. That's one of the things I dislike about matplotlib (they use this quite a bit). 

If we go for a single Ellipsoid class, then every method could take the full set of coordinates and be done with it. This is already what we do with the coordinate conversions and would make everything consistent. Then you know that methods of Boule always take longitude, latitude, height/radius, in that order.

santisoler: In reply to @dabiged. I would try to avoid passing coordinates as `kwargs`. Mainly because this is not how the rest of the Fatiando libraries handle passing coordiantes to functions or methods. I think that will bring some awkwardness.
For example, I can create a regular grid using Verde and then use Boule to compute the normal gravity on each point of the grid. Currently I can do that through:
```python
import verde as vd
import boule as bl

coordinates = vd.grid_coordinates(region=(..., ..., ..., ...), spacing=..., extra_coords=1e3)
ellipsoid = bl.WGS84
ellipsoid.normal_gravity(*coordinates[1:])
```
I think this is much simpler than specifying each coordinate with keys.

In reply to @leouieda. I think every option has its drawbacks. I'm a little bit worried of users getting confused about how the `normal_gravity` is computed if we are asking for arguments that aren't used, but you are right we are actually doing so on coordinate conversion (although I can say that a general coordinate conversions would need the entire set of coordinates, while in this case happens that `longitude` on both systems are the same).

If we really want to keep tab-completion, I would go with your proposal and ask for the whole set of coordinates. In fact the previous snippet I wrote would be much simpler:
```python
import verde as vd
import boule as bl

coordinates = vd.grid_coordinates(region=(..., ..., ..., ...), spacing=..., extra_coords=1e3)
ellipsoid = bl.WGS84
ellipsoid.normal_gravity(*coordinates)
```

leouieda (approved):

leouieda:
@dabiged thanks for getting this started and sorry that it was stalled for so long! To push this through, I made a small fix to the docstring formatting and added a quick doctest to make sure everything was covered. Merging it in as soon as CIs all pass.

leouieda:
Ignoring the pylint failure since we're moving away from it soon anyway.

Issue #76, opened 2020-11-12
Unify the Ellipsoids into a single class


leouieda:
**Description of the desired feature**

#47 talks about implementing a factory function to instantiate ellipsoids. Recently, #73 got me thinking that we do some strange things to make sure that the `Sphere` and `Ellipsoid` classes are compatible, like asking for arguments to methods that are not used.

So how about we merge everything into a single `Ellipsoid` class? It would take 3 axis arguments (semimajor, semimedium, semiminor) with only the semimajor being mandatory. It can then dispatch the computations appropriately depending on the number of axis given (1 goes to sphere calculations, 2 goes to ellipsoid, 3 to triaxial). The calculations themselves can be moved to functions or private methods of this class. The public facing methods would all be the same method in reality so we wouldn't have to artificially make them compatible (or forget to do so). 

This would make the process of creating a new ellipsoid simpler: just use `Ellipsoid` and give it the appropriate amount of axis. It would also remove a lot of duplicate code and documentation.

**Are you willing to help implement and maintain this feature?** Yes

leouieda:
Here is what I would imagine this class would look like.

Defining attributes:

* `name`
* `geocentric_grav_const`
* `angular_velocity`
* `semimajor_axis`
* `semimedium_axis=None`
* `semiminor_axis=None`
* `flattening=None` (can be given instead of `semiminor_axis`)
* `long_name=None`
* `reference=None`

Derived attributes:

* `kind`: one of `["sphere", "triaxial", "oblate" ]` depending on the defined axis
* `mean_radius`
* `volume`
* `gravity_pole` (can be calculated independently or using `normal_gravity`)
* `gravity_equator` (can be calculated independently or using `normal_gravity`)
* `first_eccentricity` (only defined for oblate ellipsoids, others result in an exception)
* `second_eccentricity` (only defined for oblate ellipsoids, others result in an exception)
* `linear_eccentricity` (only defined for oblate ellipsoids, others result in an exception)

Public methods:

* `geocentric_radius(longitude, latitude, geodetic=True)`: distance from center to surface. For a sphere this is the `semimajor_axis`. Latitude is geocentric or geodetic (only makes a difference for oblate ellipsoids).
* `normal_gravity(longitude, latitude, height)`: the norm of the gradient of the gravity potential (gravitational + centrifugal). Latitude is geodetic 
* `normal_gravitation(longitude, latitude, height)`: the norm of the gradient of the gravitational potential
[...]

PR #77, opened 2020-11-16
Unify Spheres, oblate spheroids and triaxial ellipsoids into one class.

dabiged:
This PR unifies all ellipsoids (Spheres, oblate spheroids and triaxial ellipsoids) into a single class.

For discussion see #76 

**Reminders**:


leouieda:
@dabiged thanks for starting this 🙂 I would suggest writing the tests first and then working on the implementation. We already have most of the tests for oblate ellipsoids and spheres and those should all pass with the new class. 

For this PR, it might be best to omit anything for triaxial ellipsoids (raising `NotImplementedError`). This will already be a lot of work without having to figure out new tests for the triaxial part. We can then work on that in #72 

I'm having a loot at what we can do with `attrs` to figure out the axis values from the flattening. I'll push some changes to this PR when I have a bit of time.

dabiged:
Thanks for the update. I have a job now that isn't in the geophysics space so I will see when I can get back to this!

PR #104, opened 2022-05-12
Fix the format of some files 

aguspesce:
I ran `make format` and it changed the power operators on some files according to the new version of Black


leouieda:
🎉 thanks!

PR #112, opened 2022-06-13
Use jupyter-sphinx and update front page

leouieda:
**Description**

Modernize the front page and installing guide to match the latest version of Pooch. Uses `sphinx-design` instead since it's meant to replace `sphinx-panels` in the future. Move the tutorials to `jupyer-sphinx` instead of sphinx-gallery since we're not building galleries anyway. Added a glossary that can be populated with definitions afterwards.

**Relevant issues/PRs**


leouieda:
I'll merge this in quickly to iterate but PRs and issues with feedback are more than welcome.

PR #116, opened 2022-06-14
Remove the boule.test function

leouieda:
**Description**

It doesn't really serve a good purpose and can be replaced entirely with
`pytest --pyargs boule` which is what we actually use to run the tests.
Removing this outright since it's very unlikely to be missed and the
impact is very small.

**Relevant issues/PRs**

See fatiando/community#44


leouieda:
Pinging @santisoler here. Taking a bit of a wild approach here. This should be safe enough and if no one complains maybe we should do the same to Harmonica?

PR #119, opened 2022-06-14
Make all package modules private

leouieda:
**Description**

The modules inside the package are mostly a way for us to organize code.
Users are meant to use things from the top-level namespace which we
define in `boule/__init__.py`. To avoid confusion and the modules
showing up in tab-completion, make them all private by adding a leading
`_`.

**Relevant issues/PRs**


leouieda:
Merging this in since it's an internal change that ideally shouldn't impact users.

Issue #31, opened 2020-04-15
Point to Harmonica's Gravity Disturbances gallery example on Normal Gravity


santisoler:
**Description of the desired feature**

On the Normal Gravity gallery example, would be nice to point towards the [Harmonica's Gravity Disturbances](https://www.fatiando.org/harmonica/latest/gallery/gravity_disturbance.html) gallery example, just to show how the `normal_gravity` method could be applied to the computation of the gravity disturbances.

The lack of clearness on how `normal_gravity` should be used was raised by @craigmillernz on fatiando/harmonica#164.

**Are you willing to help implement and maintain this feature?** Yes

santisoler:
As discussed in last meeting, I agree with moving the gravity disturbance tutorial to Boule. It also sounded weird that the first two pages of Harmonica User Guide don't use Harmonica at all. And the same goes for the coordinates systems: since it's something we use across the Fatiando libraries, it should be placed in a different place than Harmonica's docs.

leouieda:
Closing this since there is a tutorial here already.

PR #130, opened 2022-06-24
Refactor the documentation tutorials

leouieda:
**Description**

Shifted the focus to how to do things with Boule and point to other pages (like the glossary and docstrings) for more of the theory. Gravity disturbance tutorial expanded a bit to be more in line with the Harmonica tutorial. Coordinate conversion tutorial was refocused on using pymap3d instead of explaining what the coordinate systems are, which should be in a separate page. 

**Relevant issues/PRs**

See #31 for discussion on disturbance and coordinate conversion tutorials.


leouieda:
@LL-Geo @aguspesce @santisoler @MGomezN I could really use a few eyes on this 👀. What do you think of the changes? I do want to keep the extensive written explanations about the coordinate systems but I'm not sure where to put it. See #31 as well.

leouieda:
Actually, I'll do as @santisoler did for Harmonica and merge these in so we can see the results in https://www.fatiando.org/boule/dev/ and make improvements/changes in follow-up PRs.

PR #136, opened 2022-09-02
Add serve target in doc/Makefile

santisoler:
Add a new `serve` target in `doc/Makefile` so it's easy to serve the docs locally.


leouieda (approved):
Thanks for this @santisoler!

PR #139, opened 2022-09-23
Add definition of “co-located grids” to the glossary

### LINKED ISSUE #138: Add meaning of "co-located"
As an absolute beginner working with xarray and non-native English, the word "co-located" in this sentence was unclear:
 "Now we can download and open co-located topography and geoid grids."
I had to read Ensaio documentation to understand both data sets have node spacing of 10 arc-minutes, the horizontal datum is WGS84 and there are 1081 x 2161 grid points in total. I conclude this is the meaning of "co-located".
Can we explicitly say what co-located means and guide people working with different geoids/data sets on the requirements of the grids? I mean, direct people to some external tutorial where they learn how to "co-locate" grids. 
Thanks!

MGomezN:
Addition requested in #138

**Relevant issues/PRs:**

Fixes #138

### DIFF
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -5,8 +5,11 @@ Glossary
 
 .. glossary::
 
-    Co-located grids.
-        The grid points are at exactly the same locations. It's important because otherwise we can't add or subtract values since they wouldn't be aligned. See `Normal gravity <https://www.fatiando.org/boule/latest/overview.html#normal-gravity>`__.
+    Co-located grids
+        One or more grids that have grid points are at exactly the same locations. 
+        This property is important when performing operations like adding and 
+        subtracting grid values since they must be located at the same 
+        geographical locations.
         
     Geocentric gravitational constant
         The multiplication of the total mass of a :term:`reference ellipsoid`
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -5,27 +5,38 @@ Glossary
 
 .. glossary::
 
-    Reference ellipsoid
-        An ellipsoid with geometric and physical properties used to best
-        approaximate a given planetary body (the Earth or others).
-        See `Wikipedia <https://en.wikipedia.org/wiki/Reference_ellipsoid>`__.
-
+    Co-located grids.
+        The grid points are at exactly the same locations. It's important because otherwise we can't add or subtract values since they wouldn't be aligned. See `Normal gravity <https://www.fatiando.org/boule/latest/overview.html#normal-gravity>`__.
+        
     Geocentric gravitational constant
         The multiplication of the total mass of a :term:`reference ellipsoid`
         and the `gravitational constant
         <https://en.wikipedia.org/wiki/Gravitational_constant>`__. Often
         represented as :math:`GM`.
-
-    Normal gravity
-        The magnitude of the gradient of the :term:`gravity potential` of a
-        :term:`reference ellipsoid`.
-
-    Gravity potential
-        The combined gravitational and centrifugal potential
-        :math:`U = V + \Phi` [HofmannWellenhofMoritz2006]_.
-
+        
     Gravity disturbance
         The gravity disturbance is the difference between the magnitude of
         observed gravity and the magnitude of normal gravity evaluated at the
         same point [HofmannWellenhofMoritz2006]_:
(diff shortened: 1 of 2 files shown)


doc/glossary.rst
```diff
@@ -5,27 +5,38 @@ Glossary
 
 .. glossary::
 
-    Reference ellipsoid
-        An ellipsoid with geometric and physical properties used to best
-        approaximate a given planetary body (the Earth or others).
-        See `Wikipedia <https://en.wikipedia.org/wiki/Reference_ellipsoid>`__.
-
+    Co-located grids.
+        The grid points are at exactly the same locations. It's important because otherwise we can't add or subtract values since they wouldn't be aligned. See `Normal gravity <https://www.fatiando.org/boule/latest/overview.html#normal-gravity>`__.
```
leouieda, line 9:
```suggestion
    Co-located grids
        One or more grids that have grid points are at exactly the same locations. 
        This property is important when performing operations like adding and 
        subtracting grid values since they must be located at the same 
        geographical locations.
```

leouieda: @MGomezN the linking for glossary items usually goes the other way around. Instead of linking to the tutorial here, in the tutorial you can substitute the occurrence of "co-located grids" with `:term:`co-located grids` and it will automatically link to this glossary entry.

leouieda (commented):
Thanks for adding this @MGomezN ❤️ 

I have a small suggestion for rewording and linking this back to the tutorial.

doc/glossary.rst
MGomezN: Thanks for your patience and kind suggestions!

leouieda:
Pulled this commit into #139 since they make more sense as a single change.

leouieda:
Oops, just noticed that I posted messages to the wrong PRs...

leouieda:
@MGomezN this PR and #140 are better as a single change (adding to the glossary and then using the glossary term in the docs). So I pulled the commit from that PR into this one. So I'll be closing that one and commenting here instead.

In the future, you can keep pushing to the same branch to update the PR if there are further changes you want to make before the PR is merged. The balance of what goes into a single PR and what should be multiple is tricky and we've been playing it by ear. A good rule of thumb is: does this PR make sense on its own? If not, then maybe further changes can be added there. Otherwise, then it would be better as a follow-up PR. 

None of this is too strict and it's always possible to change as we go along (pulling commits into separate PRs is doable with a little git magic 😉). So don't worry too much about getting it right from the start. We rarely do and often have discussions along these lines.

doc/glossary.rst
```diff
@@ -5,27 +5,41 @@ Glossary
...
+    Reference ellipsoid
+        An ellipsoid with geometric and physical properties used to best
+        approaximate a given planetary body (the Earth or others).
+        See `Wikipedia <https://en.wikipedia.org/wiki/Reference_ellipsoid>`__.
+
+
+
+
+
+
```
leouieda, line 43:
```suggestion
    Reference ellipsoid
        An ellipsoid with geometric and physical properties used to best
        approximate a given planetary body (the Earth or others).
        See `Wikipedia <https://en.wikipedia.org/wiki/Reference_ellipsoid>`__.
```

doc/overview.rst
```diff
@@ -71,7 +71,7 @@ First, we need to import a few other packages:
     import pygmt         # For plotting maps
     import xarray as xr  # For manipulating grids
 
-Now we can download and open co-located topography and geoid grids using
+Now we can download and open :co-located:`co-located grids` topography and geoid grids using
 :mod:`ensaio` and :mod:`xarray`:
```
leouieda, line 75:
```suggestion
Now we can download and open :term:`co-located grids` of topography and geoid
using :mod:`ensaio` and :mod:`xarray`:
```

leouieda: The command for linking to the glossary is `:term:`. Using that requires rephrasing a bit to avoid the duplicated "grids".

leouieda (approved):
@MGomezN left a few suggestions which I'll merge into this PR so we can get the PR merged as well 🙂 Sorry for the delay on this.

doc/overview.rst
MGomezN: Thank you for explaining! Sometimes I am confused about where I have to substitute text and where is an actual command 😕

PR #140, opened 2022-10-19
Add co-located link glossary entry

MGomezN:
Substitute the occurrence of "co-located grids" with :term:`co-located grids` to automatically link to this glossary entry.
Originally @leouieda [suggestion](https://github.com/fatiando/boule/pull/139#discussion_r979169638).

**Relevant issues/PRs:**

Relevant to #138 and #139

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -71,7 +71,7 @@ First, we need to import a few other packages:
     import pygmt         # For plotting maps
     import xarray as xr  # For manipulating grids
 
-Now we can download and open co-located topography and geoid grids using
+Now we can download and open :co-located:`co-located grids` topography and geoid grids using
 :mod:`ensaio` and :mod:`xarray`:
 
 .. jupyter-execute::


leouieda:
@MGomezN this PR and #140 are better as a single change (adding to the glossary and then using the glossary term in the docs). So I pulled the commit from that PR into this one. So I'll be closing that one and commenting here instead.

In the future, you can keep pushing to the same branch to update the PR if there are further changes you want to make before the PR is merged. The balance of what goes into a single PR and what should be multiple is tricky and we've been playing it by ear. A good rule of thumb is: does this PR make sense on its own? If not, then maybe further changes can be added there. Otherwise, then it would be better as a follow-up PR. 

None of this is too strict and it's always possible to change as we go along (pulling commits into separate PRs is doable with a little git magic 😉). So don't worry too much about getting it right from the start. We rarely do and often have discussions along these lines.

PR #142, opened 2022-10-26
Undo deprecation of coordinate conversion methods

santisoler:
Revert "Deprecate the coordinate conversion methods (#126)" to ditch the
deprecation warnings for the coordinate conversion methods.
Restore the coordinate conversion methods using the `pymap3d` functions under
the hood. Pass the longitude and latitude arguments and get the returns in the
proper order. Add `pymap3d` as a required dependency of Boule.

**Relevant issues/PRs:**

#126, https://github.com/fatiando/harmonica/pull/350,
[Fatiando Dev Meeting of 2022-09-28](https://github.com/fatiando/community/blob/main/development-calls/2022.md#2022-09-28)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,9 +39,9 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, macos, windows]
-        python: ["3.6", "3.10"]
+        python: ["3.7", "3.10"]
         include:
-          - python: "3.6"
+          - python: "3.7"
             dependencies: oldest
           - python: "3.10"
             dependencies: latest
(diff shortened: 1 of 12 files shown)


leouieda:
@santisoler after looking this over a bit, I thought it was best to not add pymap3d as a direct dependency. It's not a lot of code, we already have in the code base + tests, and pymap3d may decide to change their ellipsoid. So let's keep this in here for convenience but make it possible to do other conversions with pymap3d.

We may have to revisit this when we do layered ellipsoids since we'd need to the Cartesian conversion as well. But that can come later.

I pushed an update to the tutorial as well to use our methods and then have a section for pymap3d.

leouieda:
Need to wait for #144 since pymap3d>2.9 doesn't exist for 3.6.

leouieda:
Also best to not add a new dependency during a bug fix release.

PR #146, opened 2023-01-16
Add geocentric_radius to TriaxialEllipsoid

### LINKED ISSUE #105: Add geocentric surface radius calculation to TriaxialEllipsoid

#72 added a `TriaxialEllipsoid` class defining only the base geometric properties. In #47 @MarkWieczorek mentioned that calculating the surface radius of the ellipsoid would be useful. Seems like a good place to start adding some computations.

https://en.wikipedia.org/wiki/Ellipsoid is a good place to start with equations and references.

Willing to help if anyone wants to give this a shot but probably won't do it myself.

blazej-bucha:
Add a new `geocentric_radius` method to `TriaxialEllipsoid` that computes the geocentric surface radius of points in triaxial ellipsoids. Add two new flattening properties to the class: the `meridional_flattening` and the `equatorial_flattening`. Add tests and references for the equations used.

---

* Add the `geocentric_radius` method to the `TriaxialEllipsoid` class that computes the radius of surface points on triaxial ellipsoids.  
* Add tests to check `geocentric_radius` of `TriaxialEllipsoid`.
* Running `make format` made few edits also to the previous code.

Please take a look at the PR to check if it makes sense.

**Relevant issues/PRs:** 

Fixes #105


santisoler:
Hi @blazej-bucha! Thanks so much for opening this PR! I'm really sorry for the delay of my reply.

I think this is a great addition. Let me review it a little bit and will come back with comments.

Regarding the code-style Action that fails, apparently you run an older version of Black that used to format every power operator (`**`) with leading and trailing spaces. The updated version of Black we use in Boule formats simple power operators without spaces between the base and the power (check [this section in Black's docs](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-breaks-binary-operators)).

boule/_triaxialellipsoid.py
```diff
@@ -184,3 +184,100 @@ def volume(self):
...
+        longitude_semimajor_axis_rad = np.radians(longitude_semimajor_axis)
+
+        coslat, sinlat = np.cos(latitude_rad), np.sin(latitude_rad)
+
+        a = self.semimajor_axis
+        b = self.semimedium_axis
+        c = self.semiminor_axis
+
+        f1 = (a - c) / a
+        f2 = (a - b) / a
```
santisoler, line 271: I understand that these two are the two flattenings described in the paper. It would be nice if these two are properties of the class, just like the `thirdflattening` is defined for the `Ellipsoid`:

https://github.com/fatiando/boule/blob/aa97c274a61e3c496656719a4c7d4cfc4916af87/boule/_ellipsoid.py#L166-L175

 We can then add some docstrings explaining them a little bit better. Their names could be `equatorial_flattening` and `meridional_flattening`. I'm open to suggestions.

boule/_triaxialellipsoid.py
```diff
@@ -184,3 +184,100 @@ def volume(self):
...
+
+        radius = (a * (1.0 - f1) * (1.0 - f2)) / np.sqrt(
+            1.0
+            - (2.0 * f1 - f1**2) * coslat**2
+            - (2.0 * f2 - f2**2) * sinlat**2
+            - (1.0 - f1) ** 2
+            * (2.0 * f2 - f2**2)
+            * coslat**2
+            * np.cos(longitude_rad - longitude_semimajor_axis_rad) ** 2
+        )
```
santisoler, line 281: From comparing with the referenced paper, I understand that `f1` is the flattening of the meridional section of the triaxial ellipsoid in whose plane lies the semi-major axis (what a mouthful), and `f2` is the equatorial flattening.

So, following the paper's notation, `f1` is equal to $\alpha$ and `f2` is equal to $\alpha_1$, am I right?

If that's so, I think the two lines that contain the `coslat` and `sinlat` might be wrong. Shouldn't it be like this:
```suggestion
        radius = (a * (1.0 - f1) * (1.0 - f2)) / np.sqrt(
            1.0
            - (2.0 * f1 - f1**2) * sinlat**2
            - (2.0 * f2 - f2**2) * coslat**2
            - (1.0 - f1) ** 2
            * (2.0 * f2 - f2**2)
            * coslat**2
            * np.cos(longitude_rad - longitude_semimajor_axis_rad) ** 2
        )
```

I'm comparing with the first equation in the paper, where these two terms are written as:

$$ - (2\alpha - \alpha^2) \sin^2 \theta - (2\alpha_1 - \alpha_1^2) \cos^2 \theta $$

Maybe I'm missing something, so please correct me if I'm wrong.

boule/tests/test_triaxialellipsoid.py
```diff
@@ -199,3 +200,85 @@ def test_volume_lt_majorsphere(triaxialellipsoid):
...
+    npt.assert_allclose(
+        radius_true, triaxialellipsoid.geocentric_radius(latitude, longitude)
+    )
+
+
+def test_geocentric_radius_equator(triaxialellipsoid):
+    """
+    Check against values at the equator
+    """
+    latitude = np.full((4,), 0.0)
```
santisoler, line 221: Minor change, this could be done with
```suggestion
    latitude = np.zeros(4)
```

santisoler (changes requested):
This looks great @blazej-bucha! I left some minor comments and suggestions. Feel free to comment them!

Looking forward to see this merged!

boule/_triaxialellipsoid.py
santisoler: Nevermind. They use the colatitude and we are using the latitude. Please, ignore this comment.

blazej-bucha:
Hi @santisoler,

Thanks for looking at the pull request and for your comments!  All your suggestions should now be applied.  In addition, reverting the order of input parameters in  `geocentric_radius` necessitated a few more edits in tests that call this method.

Formating should now hopefully be OK.  Previously, I used indeed an older version of Black from Debian repos.

santisoler (changes requested):
Thanks @blazej-bucha for addressing my comments. I know that changing the order of `longitude` and `latitude` involved a lot of changes on the testing functions. Thanks for that.

I think this is ready to be merged. I just left a few edits to the docstrings, just to improve the documentation.

Thanks again for your collaboration!

blazej-bucha:
@santisoler, thanks for your edits!  I directly commited all of them.

I'm now happy with the pull request, too.  Thanks again for reviewing the code.

santisoler (approved):

santisoler:
Thanks @blazej-bucha for the contribution! I've just made a few style changes and removed some unused variables that flake was complaining about. I'm merging this! 🚀

leouieda:
@blazej-bucha thanks for this! Please see the message above ☝🏾 about adding yourself to the `AUTHORS.md` file.

PR #147, opened 2023-03-31
Extend support for Python 3.11

santisoler:
Extend support for Python 3.11 and bump versions of GitHub Actions.


leouieda:
Thanks @santisoler

PR #173, opened 2024-03-25
Add mass, mean_density, and volume_equivalent_radius properties

### LINKED ISSUE #164: Add attributes: mass, mean_density, and volume_equivalent_radius
In addition to the attributes that are already computed for each of the ellipsoid classes, I propose that we also provide the following:
* mass (we will need to define G somewhere)
* volume_equivalent_radius 
* mean_density

MarkWieczorek:
This PR addresses the issue https://github.com/fatiando/boule/issues/164 concerning the addition of `mass`, `mean_density`, and `volume_equivalent_radius` properties to the ellipsoid classes. The following changes were made:

* Created a top level file `_constants.py` to house to the gravitational constant `G` that is imported into the three ellipsoid classes.
* Added the properties `mass`, `mean_density`, and `volume_equivalent_radius` to each of the three ellipsoid classes.
* Clarified the docstrings of `mean radius`: This is, in fact, *not* the mean radius of the sphere (i.e., the degree 0 coefficient of the shape), but rather is the mean of the semi-axes (a + b + c)/3.  This is defined as R_1 in WGS84 Moritz1988.

Comments

* The mathematical equation for the `volume_equivalent_radius` is `(self.volume * 3 / 4 / np.pi)**(1 / 3)`. If we expanded the definition of `self.volume` we could probably save a couple multiplications (but at the expense of readability).
* I don't know if we need to say "Added for compatibility with pymap3d." in the `Sphere` docstrings. For me, these quantities are defined so that all classes take the same parameters.

And, I'm happy to modify anything to conform to your coding standards that are higher than mine!

**Relevant issues/PRs:**
Fixes #164

### DIFF
--- a/boule/_sphere.py
+++ b/boule/_sphere.py
@@ -215,7 +215,6 @@ class Sphere:
     def volume_equivalent_radius(self):
         r"""
         The volume equivalent radius of the sphere is equal to its radius.
-        Added for compatibility with pymap3d.
         Definition: :math:`R_3 = R`.
         Units: :math:`m`.
         """
--- a/boule/_sphere.py
+++ b/boule/_sphere.py
@@ -181,9 +181,9 @@ class Sphere:
     @property
     def mean_radius(self):
         """
-        The mean radius of the sphere is by definition equal to its radius.
-        Added for compatibility with pymap3d.
-        Definition: :math:`R = R`.
+        The arithmetic mean radius of the ellipsoid semi-axes is equal to its
+        radius. Added for compatibility with pymap3d.
+        Definition: :math:`R_1 = R`.
         Units: :math:`m`.
         """
         return self.radius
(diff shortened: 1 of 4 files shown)


boule/_sphere.py
```diff
@@ -185,6 +193,34 @@ def volume(self):
...
+        Definition: :math:`\rho = M / V`.
+        Units: :math:`kg / m^3`.
+        """
+        return self.mass / self.volume
+
+    @property
+    def volume_equivalent_radius(self):
+        r"""
+        The volume equivalent radius of the sphere is equal to its radius.
+        Added for compatibility with pymap3d.
```
leouieda, line 218:
```suggestion
```

leouieda (approved):
Thanks for this @MarkWieczorek! Merging it in.

leouieda:
I agree that we don't need to worry about multiplying a few more times. This is not something called in a loop.

PR #175, opened 2024-03-27
Add option to use geocentric latitude with Ellipsoid.normal_gravity

### LINKED ISSUE #163: Add Ellipsoid.normal_gravitation() method
The Sphere class has the option of computing either the normal gravity or normal gravitation. In contrast, the Ellipsoid class only gives the option to compute the normal gravity. I propose that we allow the computation of the normal gravitation of an ellipsoid by creating the method `Ellipsoid.normal_gravitation`. This would help make the two classes consistent. Furthermore, there are cases (such as when you are not attached to the Earth) when the normal gravitation would be more appropriate.

MarkWieczorek:
This PR adds the option to use geocentric latitude with `Ellipsoid.normal_gravity()`.

The following changes were made:

1. Added the optional argument `geodetic` to the method `normal_gravity`.
2. When False, the geocentric latitude is converted to geodetic latitude using the exact same code as in the method `spherical_to_geodetic`. The only difference is that it is first necessary to compute the radius, and we don't care about the longitude and height parameters.
3. The documention for the routine has been modified, mostly in a subtle manner. A note was added stating that if you already have the geodetic latitude, that you should use it (otherwise there is a useless conversion).

I also made a very minor change in the doc string for geocentric_radius: "geocentric geodetic" was changed to just "geodetic" (consistent with other routines). I also note that the normal gravity doc string referred to "geometric" height, but this should be either ellipsoidal height or geodetic height: I  mentioned explicitly  that this is normal to the ellipsoid, as its meaning could be misinterpreted when using geocentric latitude.

I have tried my best to make the documentation as clear as possible, but there is probably room for improvement.

**Relevant issues/PRs:**
resolves #163


MarkWieczorek:
On further reflection, I'm not sure if this is the best approach. At this point, one can enter either 
* geodetic latitude and geodetic height, or
* geocentric latitude and geodetic height. 
The second might seem a bit odd.

A different way to handle this would be to allow to enter the following:
* geodetic latitude and geodetic height, or
* geocentric latitude and geocentric height (i.e., height above the ellipsoid along the vector directed to the ellipsoid center), or
* ellipsoidal-harmonic latitude and u (the semiminor axis of the ellipsoid that passes through the point).

We could then tell the normal gravity routine which coordinates we are using by specifying an optional parameter like `coords` or `coordinates` that could take values of "geodetic" (default), "geocentric", or "ellipsoidal" / "ellipsoidal-harmonic".

leouieda:
I'm in favor of adding a `coordinate_system` argument instead of `geodetic`. The docstring can mention that the calculations are done in the geodetic system so can be more efficient to do the conversions outside the method. The options could be `"geodetic"`, `"spherical"` and `"ellipsoidal-harmonic"` (since ellipsoidal coordinates also exist and are slightly different). I'd prefer to use spherical instead of geocentric since by definition the geodetic system is also geocentric. But I'm open to discussion if you disagree 🙂

leouieda:
@MarkWieczorek coming back to Boule after a while and I wanted to see if we can wrap this up. How about this:

We change all gravity calculation functions take a `coordinates` and a `coordinate_system` argument. The `coordinates` is a tuple of 3 arrays. What they are depends on `coordinate_system`. If it's `geodetic`, then the 3 arrays are `longitude, latitude, height`. If `geocentric` then `longitude, spherical_latitude, radius`. If `ellipsoidal harmonic`, then `longitude, beta, u`. I'm proposing using the argument `coordinates` so we avoid a situation with `latitude=None, height=None, radius=None, ...`. 

Then we handle conversions between coordinate systems internally. For spherical, we'd have to go from spherical to geodetic to ellipsoidal harmonic (all using methods we already have). We can then add to the docstrings saying that equations are naturally in ellipsoidal harmonic so that will be the most efficient and to do conversions independently if running gravity calculations in a loop. 

This would be **backward incompatible** since we'd change the signature of the gravity functions. To be honest, I was already a bit bothered that they didn't take `longitude` in the first place for consistency of the API. So I'm OK with this change breaking things. 

What do you think?

cc @santisoler for another opinion.

PR #178, opened 2024-03-31
Add area and area_equivalent_radius to all three ellipsoid classes

MarkWieczorek:
This PR adds the property `area` and `area_equivalent_radius` to the three ellipsoid classes. The "area_equivalent_radius" is what WGS84 calls $R_2$.

The equations for the area of bi-axial and tri-axial ellipsoids are taken from https://en.wikipedia.org/wiki/Ellipsoid#Surface_area. The equation for a bi-axial ellispoid is relatively simple, and only requires to use `np.arctanh`. The equation for a triaxial ellipsoid, however, requires computing two elliptic integrals. To do this, I have used the elliptic integral functions from scipy: `scipy.special.elliprd` and `scipy.special.elliprf`.  This required adding scipy to the environment.yml and requirements-test.txt files.

I have verified that the area equivalent radius for WGS84 in boule  is what is given in the literature. I have verified that the triaxial ellipsoid routine gives the correct value when a=b for WGS84.

If you don't like having scipy as a dependency, we could perhaps have this an an optional dependency, and check if it is available before running TriaxialEllipsoid.area (this is the only routine that requires it).

**Relevant issues/PRs:**

### DIFF
--- a/boule/_triaxialellipsoid.py
+++ b/boule/_triaxialellipsoid.py
@@ -11,7 +11,7 @@ from warnings import warn
 
 import attr
 import numpy as np
-from scipy.special import elliprd, elliprf
+import scipy.special
 
 from ._constants import G
 
@@ -231,19 +231,20 @@ class TriaxialEllipsoid:
     def area(self):
         r"""
         The area of the ellipsoid.
-        Definition: :math:`A`.
+        Definition: :math:`A = 3 V R_G(a^{-2}, b^{-2}, c^{-2})`, in which
+        :math:`R_G` is the completely-symmetric elliptic integral of the second
+        kind.
         Units: :math:`m^2`.
         """
         # see https://en.wikipedia.org/wiki/Ellipsoid#Surface_area
-        a = self.semimajor_axis
-        b = self.semimedium_axis
-        c = self.semiminor_axis
-        return 2 * np.pi * c**2 + 2 * np.pi * a * b * (
-            elliprf((c / a) ** 2, (c / b) ** 2, 1)
-            - (1 / 3)
-            * (1 - (c / a) ** 2)
-            * (1 - (c / b) ** 2)
-            * elliprd((c / a) ** 2, (c / b) ** 2, 1)
+        return (
+            3
+            * self.volume
+            * scipy.special.elliprg(
+                1 / self.semimajor_axis**2,
+                1 / self.semimedium_axis**2,
+                1 / self.semiminor_axis**2,
+            )
         )
 
     @property
--- a/boule/_triaxialellipsoid.py
+++ b/boule/_triaxialellipsoid.py
@@ -100,6 +100,8 @@ class TriaxialEllipsoid:
     parameters:
 
     >>> print(f"{ellipsoid.mean_radius:.0f} m")
+    259813 m
+    >>> print(f"{ellipsoid.semiaxes_mean_radius:.0f} m")
     262700 m
     >>> print(f"{ellipsoid.area:.10e} m²")
     8.6562393883e+11 m²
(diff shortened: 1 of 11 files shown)


MarkWieczorek:
Please note that the python 3.7 tests are failing only because the scipy elliptic integral functions I used  were introduced in version 1.0.8, and that version of scipy requires python 3.8+.

See https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.elliprd.html#scipy.special.elliprd and https://docs.scipy.org/doc/scipy/release/1.8.0-notes.html.

MarkWieczorek:
Note: I have updated with github tests to use Python 3.8 and 3.12.

leouieda:
I opened #188 to drop 3.7 support than we can merge that into here once it's on main. Thanks!

env/requirements-test.txt
```diff
@@ -3,3 +3,4 @@ pytest
 pytest-cov
 coverage
 pymap3d>=2.9.0
+scipy
```
leouieda, line 6: With this, Scipy is now a runtime requirement so it should be added to `pyproject.toml` instead of here. When you do, please set the lower version boundary to `>=1.0.8` since that's the minimum we need, right?

boule/_triaxialellipsoid.py
```diff
@@ -100,6 +101,10 @@ class TriaxialEllipsoid:
 
     >>> print(f"{ellipsoid.mean_radius:.0f} m")
     262700 m
+    >>> print(f"{ellipsoid.area:.10e} m²")
+    8.6562393883e+11 m²
+    >>> print(f"{ellipsoid.area_equivalent_radius:0.0f} m")
+    262458 m
```
leouieda, line 107: @MarkWieczorek out of curiosity, where did these numbers come from?

leouieda (commented):
@MarkWieczorek thanks for this! Just a minor change to add Scipy as a runtime dependency and then resolving the conflict with the workflow file. After that, I'll merge right away.

boule/_triaxialellipsoid.py
MarkWieczorek: That number comes from the output of the boule command.

env/requirements-test.txt
MarkWieczorek: I added scipy to pyroject.toml and removed it from requirements-test.txt

MarkWieczorek: and the minimum version is 1.8.0, not 1.0.8 as I stated earlier...

leouieda: 👍🏾

boule/_triaxialellipsoid.py
```diff
@@ -222,6 +227,25 @@ def semiaxes_mean_radius(self):
...
+        a = self.semimajor_axis
+        b = self.semimedium_axis
+        c = self.semiminor_axis
+        return 2 * np.pi * c**2 + 2 * np.pi * a * b * (
+            elliprf((c / a) ** 2, (c / b) ** 2, 1)
+            - (1 / 3)
+            * (1 - (c / a) ** 2)
+            * (1 - (c / b) ** 2)
+            * elliprd((c / a) ** 2, (c / b) ** 2, 1)
+        )
```
leouieda, line 247: @MarkWieczorek I was wondering how you arrived at this expression from the one in the Wikipedia link. Scipy does have the incomplete elliptic integrals of first and second kind. So why use this version?

MarkWieczorek: I hate citing wikipedia. I also hate using it as a primary source like this, but it is only the area, so what could go wrong?

So if you go one edit earlier, you find the above equation: https://en.wikipedia.org/w/index.php?title=Ellipsoid&oldid=1219229944

Then there was an edit yesterday that says "Use simpler expression of the area in terms of Carlson symmetric integrals." The new expression only uses elliprg, instead of elliprf and elliprd. The new equation looks more elegant, so like me change it and check that we get the same answer.

MarkWieczorek: As to why use that expression, the old wiki page stated "Unlike the expression with F(φ, k) and E(φ, k), the variant based on the Carlson symmetric integrals yields valid results for a sphere and only the axis c must be the smallest, the order between the two larger axes, a and b can be arbitrary."

leouieda: 😆 yeah I know what you mean! Thanks for the explanation! I changed it to use the RG version and tests pass. I'll push that here and merge.

MarkWieczorek: You beat me to it. It turns out that the order of the three arguments to the elliprg function doesn't matter, so this is much better.

leouieda:
Merged! Many thanks for this!

Issue #176, opened 2024-03-28
New ellipsoids : Proposal


MarkWieczorek:
I would like to add a large number of ellipsoids to Boule. These would include: Mercury, Venus, Moon, Mars, Ceres, Vesta, Eros, Io, Europa, Ganymede, Callisto, Enceladus and Titan. Some of these would update the ones already present in Boule. The number of ellipsoids will likely increase over time, and many of the ellipsoids will be updated by future measurements. The current manner that Boule deals with ellipsoid realizations in not adapted to this large increase in number.

There are two issues: How do we store and organize these in Boule, and what do we call them?

## Organization
In order to make it easy to find the raw ellipsoid declarations, I propose that we convert the file `_realizations.py` into several: one for each planet, the Moon, and asteroids
* `_mercury.py`
* `_venus.py`
* `_earth.py`
* `_moon.py` (this could maybe incorporated into earth)
* `_mars.py`
* `_asteroids.py`
* `_jupiter.py` (all moons of Jupiter)
* `_saturn.py` (all moons of Saturn)

(more to be added as necessary...)

The realizations in these files would then be imported into a module called "ellipsoids" (i.e., `boule.ellipsoids`). Initially, I think that we can just import them all into one space. However, if we start adding a large number of small moons of the outer planets, or if we start getting a lot of versions for a single body, we might want to consider organizing them into submodules, like `ellipsoids.Jupiter.Ganymede`.

A question is whether we should put all these files in the top level directory, or put them in a subdirectory.

## Naming scheme
It is important to have some kind of "version control" for the ellipsoids that we propose. The parameters might change slightly over the years with improved ephemerides, or we might get new data from spacecraft. It is imortant that we don't change the values in an ellipsoids without telling people!

I propose a simple naming scheme that is just the body name, and the year that we defined the ellipsoid. If two ellipsoids are generated in the same year, we could add "a", "b, "c", For example `Mercury2015`, `Mercury2024`, `Mercury2024b`.

For the case where there are well defined previously published ellispoids, we could break with this convention: `WGS84`, `EGM2008`, etc. The only other body where there was an ellipsoid published was Mars: Ardalan et al. 2010. For this case, I would probably just call it `Mars2010`.

## Referencing

The new ellipsoids require  several parameters (GM, a, f, omega), and thus we might have separate references for each of these values. I propose to just put them all into a long reference string like 

```
"a, f: some reference 2010; GM: another reference 2009; omega: and yet another reference 2000"
```

[...]

PR #180, opened 2024-04-02
New ellipsoids

MarkWieczorek:
The PR contains a large number of new ellipsoids, renames a few preexisting ellipoids, and removes the prexisting Vesta ellipsoid.

Here is the list of all the implemented ellipsoids with some notes:

## Mercury
* Mercury2015 (previously MERCURY, updated the last digit of GM that was incorrect)
* Mercury2024 (new, updated R and GM)

## Venus
* Venus2015 (previously VENUS)

## Earth
* GRS80
* WGS84
* EGM96 (new)

## Moon
* Moon2015 (previously MOON)

## Asteroids
* Ceres2018 (new)
* Vesta2017 (new)
* Vesta2017_triaxial (new)
* VESTA Removed (see comments below)

## Mars
* Mars2009 (previously MARS)

## Jupiter system
* Io2024 (new)
* Europa2024 (new)
* Ganymede2024 (new)
* Callisto2024 (new)

## Saturn system
* Enceladus2024 (new)
* Titan2024 (new)

## Pluto system
* Pluto2024 (new)
* Charon2024 (new)

# Notes
I removed the original VESTA triaxial ellipsoid because there is not enough information in the paper to determine the GM, it is not stated if the ellipsoid is geocentric or offset from the center of mass, and the directions of the principal axes are not given. I think that keeping this ellipsoid as is will just cause problems for people who use it. I wasted a lot of time trying to figure this one out... Instead, two new ellipsoids have been added: The default (Vesta2017) is biaxial, as Vesta has no real tides. The Vesta2017_triaxial ellispoid, in contrast, is triaxial. The triaxial one, however, is **not** aligned with the topography and gravity coordinate axes: I think that we will probably need to add additional attributes to the TriaxialEllipsoid class for the unit vectors of the semi-axes.

Note that I have simply imported all the ellipsoids into the main namespace. I suspect that this will be too messy for people. As I mentioned in the related issue, we could instead put all of the ellipsoid realizations into a new submodule called `ellipsoids`.

In many cases, papers generated several ellipsoids, and I noted which one I chose as an inline comment. I think that these comments should be placed in an optional "comments" attribute of each class. In general, I chose to use ellipsoids that were geocentric (with their center corresponding to the center of mass), and also chose to use the "hydrostatic" constrained ellipsoids for the satellites of Jupiter.

Lastly, note that the ellipsoid year corresponds to the year when all the required data were published, or to the year that we assembled the ellipsoid ourselves using various sources.

**Relevant issues/PRs:**
Relevant to #176

### DIFF
--- a/boule/_realizations.py
+++ b/boule/_realizations.py
@@ -315,3 +315,19 @@ Pluto2024 = Sphere(
         "https://doi.org/10.1016/j.icarus.2014.03.015; "
     ),
 )
+
+Charon2024 = Sphere(
+    name="Charon2024",
+    long_name="Charon spheroid (2024)",
+    radius=606_000,
+    geocentric_grav_const=105.88e9,
+    angular_velocity=1.1385591834674098e-055,
+    reference=(
+        "R: Nimmo, et al. (2017). Mean radius and shape of Pluto and Charon "
+        "from New Horizons images. Icarus, 287, 12–29. "
+        "https://doi.org/10.1016/j.icarus.2016.06.027; "
+        "GM, OMEGA: Brozović, M., et al. (2015). The orbits and masses of "
+        "satellites of Pluto. Icarus, 246, 317–329. "
+        "https://doi.org/10.1016/j.icarus.2014.03.015; "
+    ),
+)
--- a/boule/_realizations.py
+++ b/boule/_realizations.py
@@ -171,3 +171,21 @@ Vesta2017_triaxial = TriaxialEllipsoid(
         "71–82. https://doi.org/10.1016/j.epsl.2017.07.033"
     ),
 )
+
+Ceres2018 = Ellipsoid(
+    name="Ceres2018",
+    long_name="Ceres ellipsoid (2018)",
+    semimajor_axis=482_100,
+    flattening=(482_100 - 445.94) / 482_100,
+    geocentric_grav_const=62629053612.1,
+    angular_velocity=1.9234038694078873e-4,
+    reference=(
+        "A, F: Park, R. S., et al. (2019). High-resolution shape model of "
+        "Ceres from stereophotoclinometry using Dawn Imaging Data. Icarus, "
+        "319, 812–827. https://doi.org/10.1016/j.icarus.2018.10.024; "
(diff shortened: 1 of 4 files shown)


boule/_realizations.py
```diff
@@ -7,13 +7,57 @@
...
+        "Wieczorek, MA (2015). 10.05 - Gravity and Topography of the Terrestrial "
+        "Planets, Treatise of Geophysics (Second Edition); Elsevier. "
+        "doi:10.1016/B978-0-444-53802-4.00169-X"
+    ),
+)
+
+Mercury2024 = Sphere(
+    name="Mercury2024",
+    long_name="Mercury spheroid (2023)",
+    radius=2439472.7,
```
VascoSch92, line 32: 2_439_473.7 ?

MarkWieczorek:
Please merge #183 and #185 first. Afterwards, I will make minor changes to this PR that make use of the new comments and semimajor_axis_longitude atrributes.

boule/_realizations.py
```diff
@@ -7,13 +7,57 @@
...
+    reference=(
+        "Wieczorek, MA (2015). 10.05 - Gravity and Topography of the Terrestrial "
+        "Planets, Treatise of Geophysics (Second Edition); Elsevier. "
+        "doi:10.1016/B978-0-444-53802-4.00169-X"
+    ),
+)
+
+Mercury2024 = Sphere(
+    name="Mercury2024",
+    long_name="Mercury spheroid (2023)",
```
VascoSch92, line 31: Why in the name we have 2024 and here 2023? Is that a typo? Or is it correct?

MarkWieczorek: thanks! typo...

leouieda (approved):
This is awesome! I pushed some minor fixes to the citations and renamed the Vesta triaxial to `VestaTriaxial2017` to keep the CamelCase notation.

leouieda:
> Please note that all these ellipsoids are added to the main namespace. This might be too much of a mess for people, and I suggest that we instead put these all in a new module boule.ellipsoids. I will wait for your comments before doing so.

I think this is fine. The API itself is only 3 classes so having all of these in the main namespace is OK.

PR #183, opened 2024-04-08
Turn longitude_semimajor_axis into an optional attribute

MarkWieczorek:
This PR turns `longitude_semimajor_axis` (renamed to `semimajor_axis_longitude`) into an optional attribute that is used to instantiate triaxial ellipsoids.

Previously, this parameter was found as an optional argument to the `geocentric radius` function. However, this parameter is an intrinsic property of the ellipsoid, and will be necessary for all routines that make use of longitude coordinates. As this is now declared as a class attribute, I have thus remove this optional parameter from `geocentric radius`.

The `test_geocentric_radius_semimajor_axis_longitude` was somewhat modified: I declared a new ellipsoid `triaxialellipsoid_90` that set `semimajor_axis_longitude` to 90.

As soon as this PR is merged, I will add the `centrifugal_potential` method (which requires the updated `geocentric_radius` function).

Note that at some point, we may want to allow for ellipsoids that are not rotating about the z axis, but this is probably far off into the future.

**Relevant issues/PRs:**

### DIFF
--- a/boule/_triaxialellipsoid.py
+++ b/boule/_triaxialellipsoid.py
@@ -22,8 +22,10 @@ class TriaxialEllipsoid:
 
     The ellipsoid is defined by five parameters: semimajor axis, semimedium
     axis, semiminor axis, geocentric gravitational constant, and angular
-    velocity The thee semi-axis are different and the ellipsoid spins around
-    it's largest moment of inertia.
+    velocity. The ellipsoid spins around it's smallest semiminor axis, which
+    is aligned with the Cartesian z coordinate axis. The semimajor and
+    semimedium axes are in the x-y plane, and if not specified otherwise,
+    coincide with the Cartesian x and y axes.
 
     **This class is read-only:** Input parameters and attributes cannot be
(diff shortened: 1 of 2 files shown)


boule/_triaxialellipsoid.py
```diff
@@ -22,8 +22,10 @@ class TriaxialEllipsoid:
 
     The ellipsoid is defined by five parameters: semimajor axis, semimedium
     axis, semiminor axis, geocentric gravitational constant, and angular
-    velocity The thee semi-axis are different and the ellipsoid spins around
-    it's largest moment of inertia.
+    velocity. The ellipsoid spins around it's smallest semiminor axis, which
```
VascoSch92, line 25:
```suggestion
    velocity. The ellipsoid spins around its smallest semiminor axis, which
```

MarkWieczorek: thanks. I made the change and also reworded what follows slightly.

MarkWieczorek:
This PR was incorporated into #180

Issue #151, opened 2024-02-21
Implement the normal gravity potential of the Ellipsoid


leouieda:
**Description of the desired feature:**

If we want to calculate the disturbing potential for geophysical studies, we need a way to calculate the normal gravity and gravitational potentials. Tools like [ICGEM](http://icgem.gfz-potsdam.de/calcgrid) can provide grids of the gravitational potential.

Equation 2–124 in Hofmann-Wellenhof & Moritz (2006) is the normal gravitational potential in ellipsoidal harmonic coordinates. The centrifugal term is not hard and can be found earlier in the same book. The challenge is converting the geodetic to ellipsoidal harmonic coordinates. The conversion can be found in [Lakshmanan (1991)](https://doi.org/10.1190/1.1443090). The final equation can be obtained by substitution.

The idea is to add 3 new methods to `Ellipsoid`:

1. `centrifugal_potential(self, latitude, height)`: Calculate the centrifugal potential.
2. `normal_gravitational_potential(self, latitude, height)`: Calculate the gravitational potential using the equations described above.
3. `normal_gravity_potential(self, latitude, height)`: Returns the sum of the previous 2. 

**Are you willing to help implement and maintain this feature?** 

Yes but it will take some time. Having some help would be much appreciated!

MarkWieczorek:
I think that these three functions would be very useful to have. However, what I really need for my work is simply the normal gravity potential on the surface of the ellipsoid ($U_0$), which is independent of latitude. This is simply equation 2-123 in the same book.

$U_0$ could be computed as  `normal_gravity_potential(latitude=arbitrary_value, height=0)`, but  this would be somewhat awkward if either latitude or height were required arguments. If they were both optional, we could compute it as `normal_gravity_potential()`. Perhaps a better way for my interests would be to implement $U_0$ as a property, but then we would need to come up with a unique name that is distinct from the more general function. We could call it `U0`, but this would go against the verbose naming scheme we are using.

MarkWieczorek:
For the normal gravity potential on the ellipsoid, $U_0$: 

I think that we could make this an attribute that is simply called `reference_potential`. To obtain the potential above the ellipsoid (which depends on latitude), one would then use the more general functions discussed above.

PR #184, opened 2024-04-08
Add attributes for the normal potential to Sphere and Ellipsoid

MarkWieczorek:
This PR adds the property `reference_potential` to the Ellipsoid class. This is $U_0$ as defined by eq 2-123 in Hofmann-Wellenhof and Moritz.

This quantity is required in any application where it is necessary to compute the height of the geoid above the reference ellipsoid. I verified that the value of the GRS80 ellipsoid is what is in the above book.

As part of this PR, I corrected a couple minor typos:
* The linear eccentricity was given the mathematical symbol E in the docstring (as opposed to "c" which has several other meanings)
* The volume of an ellipsoid was defined using "c" instead of "b" in the docstring.

**Note:**
I did not add `reference_potential` to the Sphere class. This is because (as noted in the docs), the surface of a rotating sphere does not correspond to an equipotential.

We could perhaps instead add
```
Sphere.reference_gravitational_potential = GM/R
```
I'm happy to add if others think this is useful. GM/R appears everywhere when working with gravity fields of bodies like the Moon, so maybe it would be useful.

**Relevant issues/PRs:**
Relevant to #151


leouieda:
Thanks for this @MarkWieczorek! Should we call this `surface_gravity_potential` instead? Makes it clear that it's not gravitational and is on the surface. Then the Sphere one could `surface_gravitational_potential`. If you find that it appears in calculations for you, then please feel free to add it.

MarkWieczorek:
Not sure which is best:

`reference_gravity_potential`
`reference_gravitational_potential`

or

`surface_gravity_potential`
`surface_gravitational_potential`

Let me think about this, unless someone want to make an executive decision.

leouieda:
No rush. I'm OK with either one.

leouieda:
😆 alright by me. And yes, it would be great to allow creating the `Ellipsoid` in multiple ways. It just makes the attribute validation logic harder but it should be doable.

PR #192, opened 2024-04-22
Add `semimedium_axis` property to sphere and ellipsoid

MarkWieczorek:
**Squash and Merge comment**

Add `semimedium_axis` and `semimajor_axis_longitude`  to the Sphere and Ellipsoid classes for compatibility with the `TriaxialEllipsoid`. Don't use mathematical symbols for the `semimedium_axis`. Remove the "Added for compatibility with pymap3d" comment in some of the properties.

---

This PR adds the property `semimedium_axis`and `semimajor_axis_longitude`  to the Sphere and Ellipsoid classes. This is done for compatibility. I note that the Sphere class already had semiminor and semimajor axes set.

Furthermore, I have removed "Added for compatibility with pymap3d" from the docstring, as I encountered this compatibility problem in pyshtools... In many cases, people will treat (Sphere, Ellipsoid, TriaxialEllipsoid) as just a boule instance, and won't care which "subclass" they are dealing with.

**Relevant issues/PRs:**
None.

### DIFF
--- a/boule/_ellipsoid.py
+++ b/boule/_ellipsoid.py
@@ -183,6 +183,15 @@ class Ellipsoid:
         return self.semimajor_axis * (1 - self.flattening)
 
     @property
+    def semimedium_axis(self):
+        """
+        The semimedium axis of the ellipsoid is equal to its semimajor axis.
+        Definition: :math:`b = a`.
+        Units: :math:`m`.
+        """
+        return self.semimajor_axis
+
+    @property
     def thirdflattening(self):
         r"""
         The third flattening of the ellipsoid (used in geodetic calculations).
--- a/boule/_ellipsoid.py
+++ b/boule/_ellipsoid.py
@@ -186,7 +186,6 @@ class Ellipsoid:
     def semimedium_axis(self):
         """
         The semimedium axis of the ellipsoid is equal to its semimajor axis.
-        Definition: :math:`b = a`.
         Units: :math:`m`.
         """
         return self.semimajor_axis
--- a/boule/_ellipsoid.py
+++ b/boule/_ellipsoid.py
@@ -192,7 +192,7 @@ class Ellipsoid:
         return self.semimajor_axis
 
     @property
-    def semimajor_axis_longitide(self):
+    def semimajor_axis_longitude(self):
         r"""
         The semimajor axis longitude of the ellipsoid is equal to zero.
         Definition: :math:`\lambda_a = 0`.
--- a/boule/_ellipsoid.py
+++ b/boule/_ellipsoid.py
@@ -192,6 +192,15 @@ class Ellipsoid:
         return self.semimajor_axis
 
     @property
+    def semimajor_axis_longitide(self):
+        r"""
+        The semimajor axis longitude of the ellipsoid is equal to zero.
+        Definition: :math:`\lambda_a = 0`.
+        Units: :math:`m`.
+        """
+        return 0
+
+    @property
     def thirdflattening(self):
         r"""
         The third flattening of the ellipsoid (used in geodetic calculations).
--- a/boule/_sphere.py
+++ b/boule/_sphere.py
@@ -153,6 +153,16 @@ class Sphere:
         return self.radius
 
     @property
+    def semimedium_axis(self):
+        """
+        The semimedium axis of the sphere is equal to its radius. Added for
+        compatibility with pymap3d.
+        Definition: :math:`a = R`.
+        Units: :math:`m`.
+        """
+        return self.radius
+
+    @property
     def semimajor_axis(self):
         """
         The semimajor axis of the sphere is equal to its radius. Added for
--- a/boule/_sphere.py
+++ b/boule/_sphere.py
@@ -170,7 +170,7 @@ class Sphere:
         return self.radius
 
     @property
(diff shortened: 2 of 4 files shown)


boule/_ellipsoid.py
```diff
@@ -182,6 +182,24 @@ def semiminor_axis(self):
         """
         return self.semimajor_axis * (1 - self.flattening)
 
+    @property
+    def semimedium_axis(self):
+        """
```
VascoSch92, line 187:
```suggestion
        r"""
```
or?

MarkWieczorek: I don't think that we need a raw string here, as there aren't any backslashes in the latex commands.

VascoSch92: ah yes you are right

boule/_ellipsoid.py
```diff
@@ -182,6 +182,24 @@ def semiminor_axis(self):
...
+        The semimedium axis of the ellipsoid is equal to its semimajor axis.
+        Definition: :math:`b = a`.
```
santisoler, line 189:
```suggestion
```

boule/_sphere.py
```diff
@@ -145,28 +145,43 @@ def _check_geocentric_grav_const(self, geocentric_grav_const, value):
...
         Definition: :math:`b = R`.
         Units: :math:`m`.
         """
         return self.radius
 
+    @property
+    def semimedium_axis(self):
+        """
+        The semimedium axis of the sphere is equal to its radius.
+        Definition: :math:`a = R`.
```
santisoler, line 158:
```suggestion
```

santisoler (commented):
Thanks @MarkWieczorek for opening this. I think this PR looks good. My only comment is related to the mathematical definition used for the semimedium axis of both the ellipsoid and the sphere. For the triaxal ellipsoid we use $a$, $b$ and $c$ for the semimajor, semimedium and semiminor axis, respectively. For the Ellipsoid and the Sphere we use $a$ and $b$ for the semimajor and semiminor, respectively.

I don't quite like seeing the definition of the semimedium axis for the Ellipsoid as $b = a$, since it may be lead to think that the semiminor axis is equal to the semimajor (because those are the symbols used in that class).

Something similar with the definition of semimedium in the Sphere. Using $a$ to define it doesn't sounds right, since that symbol is already assigned to the semimajor axis.

So, my suggestion would be to remove the definition lines, since we don't have assigned symbols for them. What do you think?

boule/_sphere.py
MarkWieczorek: Yes, this is a little confusing. For sphere, we could use `a`, `b` and `c` like this
```
    @property
    def semiminor_axis(self):
        """
        The semiminor axis of the sphere is equal to its radius.
        Definition: :math:`c = R`.
        Units: :math:`m`.
        """
        return self.radius

    @property
    def semimedium_axis(self):
        """
        The semimedium axis of the sphere is equal to its radius.
        Definition: :math:`b = R`.
        Units: :math:`m`.
        """
        return self.radius

    @property
    def semimajor_axis(self):
        """
        The semimajor axis of the sphere is equal to its radius.
        Definition: :math:`a = R`.
        Units: :math:`m`.
        """
        return self.radius
```
This would use the same notation as `TriaxialEllipsoid`. However, since sphere doesn't use any of `a`, `b`, or `c`, I am ok with just removing the definition.

For `Ellipsoid`, it is more complicated, as `semiminor_axis` is defined as `b`, which is inconsistent with the notation for `TriaxialEllipsoid`. Either we can rename the variables to use `a`, `b` and `c`, or do as you suggest and just remove the definition.

My personal preference would be to use `a`, `b` and `c` consistently in all classes, but this is only because in my work (with things like the moment of inertia and rotation axis), this is standard notation, and all bodies are triaxial. I understand that `a` and `b` are used in a lot of textbooks or the Earth's biaxial shape, so I can understand why we might want to keep it that way too.

I suggest that we do whatever you decide. The `TriaxialEllipsoid` class is too developed at this point, and most people are just going to use the `Ellipsoid` class anyways.

santisoler: I agree with your point. It does sounds weird that for Ellipsoids b is the semiminor axis and for Triaxial b is the semimedium. But I think changing the symbols we use for them is kind of out of the scope of this PR. For this PR I think removing the definitions is enough.

I'm ok to keep the definitions of `a`, `b` and `c` in the Sphere, since they don't have a particular role.

So, what if we:
- Remove the mathematical definition of the semimedium in the Ellipsoid.
- Use `a`, `b` and `c` in the Sphere for the semimajor, semimedian and semiminor axis, respectively.

MarkWieczorek:
I just made some changes using what I think is the easiest and  most consistent solution: For those mathematical symbols that are not used, I deleted them.

This means that the mathematical symbols a, b, and c where removed from Sphere, and the symbol for the semimedium axis was removed from Ellipsoid.

If we revisit this later, we'll probably have to change all the definitions in Ellipsoid to be consistent with TriaxialEllipsoid. That would involve changing a bunch of equations, so lets leave that to another PR if someone feels strongly about this!

santisoler (approved):
Looks great, @MarkWieczorek! I'm merging this.

PR #195, opened 2024-04-25
Replace sphinx napoleon for numpydoc

santisoler:
Add `numpydoc` to requirements for building the docs. Add `numpydoc` to Sphinx configuration file. Update Sphinx template for classes. Move admonitions in `Ellipsoid` and `Sphere` classes to the Notes section so `numpydoc` can render them properly.

**Relevant issues/PRs:**

Related to fatiando/community#149


santisoler:
I'm noticing that the attributes section only show a shorten description, without the mathematical definitions. I think we should fix that before merging this PR.

santisoler:
I restored the attributes list in the class template, so we can have both the attribute list generated by numpydoc (with just a single line description) and the extended descriptions below. The attributes in the list generated by numpydoc link to their corresponding extended description, making the navigation process easy.

I think I like this @leouieda, let me know what do you think.

leouieda:
I'll have a look at the built docs locally as soon as I can. Thanks for this @santisoler!

santisoler:
No worries! And also remember you can download the built docs from the Actions' artifact: https://github.com/fatiando/boule/actions/runs/9421670139

santisoler:
Update: The mathematical description is not shown in the first list of attributes, but they are shown in the extended one:

First list:
image

Extended list:
image

I think we shouldn't worry too much about this then. The mathematical expressions and the units are being shown.

leouieda (approved):
👍🏾 merge away!

Issue #194, opened 2024-04-25
Normal gravity of a sphere and the small flattening limit of an ellipsoid


MarkWieczorek:
<s>We previously had a discussion about how to define normal gravity on a sphere. At the time, I think that I was probably a little confused, but now that I see how Boule works and defines things, and after have looked through Chapter 2 of Physical Geodesy again, I think that the way we treat the Sphere class is not entirely consistent with the Ellipsoid class.

Start with the properties of a reference ellipsoid, as found in the boule documentation:

- The gravity potential (gravitation + centrifugal) is constant on the surface.
- The internal density structure is unspecified but must lead to a constant potential at the surface.

However, for a sphere, the assumptions are different:

- The internal density structure is unspecified but must be either homogeneous or vary radially (e.g., in concentric layers).
- A constant gravity potential on the surface of a rotating sphere is not possible (when the internal density structure varies only as a function of radius).

Because of this, if you take an Ellipsoid, and gradually decrease the flattening to zero, you will not asymptotically approach the results of a sphere. This is because the gravity potential is not constant on the surface of the sphere, but it is on an ellipsoid.

Now, I agree that the assumptions for the Sphere class are correct if we assume that the body is a fluid in hydrostatic equilibrium. However, we can add arbitrary density anomalies in the mantle in order to generate a gravity potential (gravitation+centrifugal) at the surface that is constant, just like in the case for Ellipsoid. The mathematical problem is that if you put `flattening=0` in the Ellipsoid equations, you will get divide by zero errors, but as I will show below, these can be avoided by taking the limit where the semiminor axis approaches the semimajor axis.

What I propose is the following: use the same assumptions for Sphere as for Ellipsoid, and then with the equations in Ellipsoid, find the asymptotic limits as the flattening goes to zero. I note that the following equations could probably have been more easily derived starting with spherical harmonics instead of ellipsoidal harmonics, but I will follow the geodesy tradition of doing things the hard way!</s>

In Boule.Ellipsoid, numerical issues are encountered when the flattening goes to zero. Here, I provide equations that are more precise in the limit where the flattening goes to zero. In fact, using these new equations, one can even use a flattening of zero in the ellipsoid class. I note that a flattening of zero in the ellipsoid class is not equivalent to a Boule.Sphere class, and I have updated the documentation to make this a little more clear. 

Though I have derived these equations myself, I provide equation numbers in the Physical Geodesy book where the equations can be verified.

# reference normal gravity potential

Start with eq. 2-123,

[...]

PR #197, opened 2024-05-03
WIP: Small flattening ellipsoids

MarkWieczorek:
*WIP*

This PR extends the Ellipsoid class to allow the use of flattenings that are close to, and including, zero. The textbook equations used in the previous version ran into numerical instabilities when the flattening was lower that about $10^{-7}$. This was a result of either divide by 0 errors when the linear eccentricity approached zero, or a result of square roots of negative numbers when computing the angle $\beta^{\prime}$. 

This PR uses Taylor series expansions about $f=0$ that do not suffer from these numerical problems. The equations and their applicability are presented in #194. In general, for each routine that is concerned, the flattening is checked, and if it is smaller than a given value, the small flattening equations are used instead of the textbook equations.

The following routines have been modified:
* `Ellipsoid.reference_normal_gravity_potential`
* `Ellipsoid.normal_gravitational_potential`
* `Ellipsoid.normal_gravity_potential`
* `Ellipsoid.normal_gravity`
* `Ellipsoid.gravity_equator`
* `Ellipsoid.gravity_pole`
* `Ellipsoid.geodetic_to_ellipsoidal_harmonic`

Note that the code in `Ellipsoid.geodetic_to_ellipsoidal_harmonic` is largely repeated in `Ellipsoid.normal_gravity`. It might be preferable to simply call `geodetic_to_ellipsoidal_harmonic` within `normal_gravity` for clarity.

**Relevant issues/PRs:**
Relevant to #194


MarkWieczorek:
After fixing a bunch of merge conflicts, this PR is ready to be reviewed. It should be trivial to verify that the case when the flatting is large corresponds to the old code. The only thing to do is to very the small flattening limits that are presented in #194.

boule/tests/test_ellipsoid.py
```diff
@@ -392,3 +375,26 @@ def test_normal_gravity_gravitational_centrifugal_potential(ellipsoid):
...
+                geocentric_grav_const=WGS84.geocentric_grav_const,
+                angular_velocity=WGS84.angular_velocity,
+            )
+            _ = ellipsoid.reference_normal_gravity_potential
+            _ = ellipsoid.gravity_pole
+            _ = ellipsoid.gravity_equator
+            _ = ellipsoid.normal_gravitational_potential(latitude, height)
+            _ = ellipsoid.normal_gravity_potential(latitude, height)
+            _ = ellipsoid.normal_gravity(latitude, height)
+        assert len(warn) == 0
```
santisoler, line 400: Small improvement to the test function by parametrizing it across the flattening values, and using the `simplefilter` to ensure that no warning is raised (reference: https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests).

```suggestion
@pytest.mark.parametrize(
    "flattening", 10.0 ** (-np.arange(1, 20, 0.5)), ids=lambda f: f"{f:.1e}"
)
def test_no_warning_small_flattenings(flattening):
    """
    Check that no warnings arise when using small values for the flattening
    """
    latitude = np.linspace(-90, 90, 13)
    height = np.array([[1.0e3] * len(latitude)])
    with warnings.catch_warnings():
        warnings.simplefilter("error")  # raise error if warning is raised
        ellipsoid = Ellipsoid(
            name="WGS84 with small flattening",
            semimajor_axis=WGS84.semimajor_axis,
            flattening=flattening,
            geocentric_grav_const=WGS84.geocentric_grav_const,
            angular_velocity=WGS84.angular_velocity,
        )
        _ = ellipsoid.reference_normal_gravity_potential
        _ = ellipsoid.gravity_pole
        _ = ellipsoid.gravity_equator
        _ = ellipsoid.normal_gravitational_potential(latitude, height)
        _ = ellipsoid.normal_gravity_potential(latitude, height)
        _ = ellipsoid.normal_gravity(latitude, height)
```

santisoler (changes requested):
Thanks @MarkWieczorek for all the hard work behind this. Solving numerical instabilities with small numbers and juggling with machine precision requires a lot of effort.

I just pushed one small change to this branch that moves the caution in the Ellipsoid docstring away from the Parameters section and puts it above it. I thought it would be easier to push the change than to suggest it in the review.

I also left two minor suggestions down below.

I think this PR is almost ready. The one thing I would add is more tests for the new cases that were introduced. Right now the 100% coverage is achieved because of the test function that ensures that no warning is being raised, but that tests is not actually ensuring that the result we get out of the new lines is the one we expect.

One way to solve this would be to run all the tests we have for the regular ellipsoids (large flattening) with another ellipsoid with zero flattening. We can quickly do this by adding a new ellipsoid with small flattening to the `ELLIPSOIDS` list in `test_ellipsoid.py`.

Moreover, I would also like to add some sort of test that ensures that the physical properties are continuous for an ellipsoid with a flattening slightly below the "small flattening" threshold and for an ellipsoid with a flattening slightly above the threshold. For example, I would expect the `gravity_equator` for an ellipsoid with `flattening=1.24e-5` to be quite close to the one with a `flattening=1.26e-5`. I can totally help with this if I'm asking too much!

Let me know what do you think!

MarkWieczorek:
The one thing that is annoying about this PR is that the flattening cutoff is somewhat arbitrary. I don't think that we can do much about that, but what we could do is check that the results for using a flattening of flattening_cutoff +/- small_value give the same result within some predefined relative precision (like 1.e-8).

If we really wanted to do even better, we could add some higher order terms to some of the small flattening limits, which should improve the accuracy if we choose the cutoff value incorrectly. I really don't want to go through all the derivations from scratch, but the Physical geodesy book does give higher order terms for some of the equations (but not all) in the associated issue : $U_0$, $V(u, \beta)$, $\gamma_a$, $\gamma_b$. I don't think that they give an expression for $U(u, /beta)$ and the two terms for the normal gravity above the ellipsoid and $\beta'$, but I'll look again.

santisoler:
I understand the motivation.

I wouldn't stress too much on deriving the higher order terms until we are sure we need those. That's why I suggested to focus on adding more tests that ensure that the approximations we have already are accurate enough.

For example, if we run this little script:
```python
import numpy as np
import boule as bl
import matplotlib.pyplot as plt

kwargs = dict(
    name="foo",
    semimajor_axis=bl.WGS84.semimajor_axis,
    geocentric_grav_const=bl.WGS84.geocentric_grav_const,
    angular_velocity=bl.WGS84.angular_velocity,
)

threshold = 1.25e-5
half_range = 0.25e-5
flattenings = np.linspace(threshold - half_range, threshold + half_range, 501)
ellipsoids = (bl.Ellipsoid(flattening=f, **kwargs) for f in flattenings)
gravity_equator = np.array([e.gravity_equator for e in ellipsoids])

plt.plot(flattenings, gravity_equator)
plt.axvline(x=threshold, linestyle="--", color="grey")
plt.xlabel("Flattening")
plt.ylabel("Gravity on the equator [m/s2]")
plt.show()
```

We get this:
Figure_1

Where the grey dotted line represents the threshold. I can still see some instabilities on flattenings larger than the threshold, so maybe we would need to increase it and see if the approximation is still valid. If not, I agree that maybe we should explore higher order terms.

PR #198, opened 2024-05-07
Run tests with oldest dependencies on x86 macos

santisoler:
Change configuration of tests in GitHub Actions: use the latest x86 macos runner with the `oldest` dependencies and Python 3.8. Use the latest macos runner (arm64) only against the `latest` and `optional` requirements. This fixes CI failing because there are not Numpy binaries in PyPI for arm64 and compatible with Python 3.8.

**Relevant issues/PRs:**

Inspired in fatiando/harmonica#499


santisoler:
Close in favor of #200

PR #207, opened 2024-08-27
Replace `build` for `python-build` in environment.yml

santisoler:
The `build` package in `conda-forge` is outdated, `python-build` should be used instead. Add step in `test.yml` GitHub Action to replace `build` for `python-build` in the collected dependencies that will be installed with conda.


santisoler:
I'm merging this so we don't block other PRs. Feel free to modify it later if you don't quite like my fix to the Actions.

